PR #3, opened 2021-02-26
Remove deps from crate documentation

bcecchinato:
Using --no-deps will remoe all transient deps in the documentation which will prevent to overload it.


nickray:
I like having my deps' docs at hand, particularly if they're patched local versions. I'd accept an additional target `docs-no-deps` that does what you prefer.

bcecchinato:
Sure I can add a new target. You might be able to use it if you plan to export the documentation on docs.rs.

nickray:
Thanks 😁

PR #4, opened 2021-03-04
Save keys with u32 header (flags+kind), followed by material.

nickray:
No more CBOR.


nickray:
Key format has been "revisited": there is now room to store additional key flags in the future, so we should not have to revisit the binary format of keys stored in flash (store::Keystore).

Next up is certificates, but this is a new PR.

PR #8, opened 2021-05-12
Get Tests Building

Eosis:
Awesome work on Trussed and Solo Keys. 😄

I cloned the repo and found the tests wouldn't build. I've made changes to fix this in the most obvious way, and verified that the tests are still passing on my machine.


nickray:
Thank you!

PR #10, opened 2021-06-01
use cfg-if to take highest client-x feature

conorpp:


nickray:
Much more convenient; thanks :)

PR #15, opened 2021-10-18
Add UiClient::wink method

robin-nitrokey:
This patch adds the UiClient::wink method that triggers a visible or
audible response of the device, allowing the user to identify it.  This
is for example needed to implement the CTAPHID wink command.


robin-nitrokey:
I’ve added default implementations for all methods of the UserInterface trait.  (Sorry if you receive this comment twice, Github apparently didn’t process my email reply.)

PR #25, opened 2022-03-15
Platform.status method

nickray:


src/service.rs
```diff
@@ -479,13 +481,15 @@ impl<P: Platform> ServiceResources<P> {
             },
 
             Request::RequestUserConsent(request) => {
-                assert_eq!(request.level, consent::Level::Normal);
+                // assert_eq!(request.level, consent::Level::Normal);
```
alistair23, line 484: Why comment this out?

src/store.rs
```diff
@@ -461,6 +462,7 @@ pub fn read<const N: usize>(store: impl Store, location: Location, path: &Path)
 }
 
 /// Writes contents to path in location of store.
+#[inline(never)]
```
alistair23, line 465: Why not just let the compiler determine this?

nickray: The compiler is bad at optimising for low stack use.

src/service.rs
nickray: There is little point in panicking in production code. If anything, this should be a debug_assert. Also we handle the Strong case too. Or do you mean I should have deleted the line completely? :)

nickray:
It seems I merged this outside of GitHub, hence closing.

src/service.rs
alistair23: I meant you should delete the line completely. There isn't much use in having commented out code

src/store.rs
alistair23: We have tooling in Tock for this. You can pass ` -Z emit-stack-sizes` and then analyse the output, like this: https://github.com/tock/tock/blob/master/boards/Makefile.common#L374

nickray: Sweet! I'll have to investigate :)

PR #27, opened 2022-06-27
Add low-level store activate; correctly set MAX_SERVICE_CLIENTS

daringer:
* `MAX_SERVICE_CLIENTS` was always set to 5, now correctly set it according to the feature provided
* add `store::activate` to allow offloading filesystem/storage initialization into the runner

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -235,6 +235,24 @@ macro_rules! store { (
         }
 
         #[allow(dead_code)]
+        pub fn init_raw(ifs: &'static littlefs2::fs::Filesystem<$Ifs>,
+                        efs: &'static littlefs2::fs::Filesystem<$Efs>,
+                        vfs: &'static littlefs2::fs::Filesystem<$Vfs>
+        )
+            -> Self
+        {
+            let store_ifs = $crate::store::Fs::new(ifs);
+            let store_efs = $crate::store::Fs::new(efs);
+            let store_vfs = $crate::store::Fs::new(vfs);
+            unsafe {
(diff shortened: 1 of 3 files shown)


src/store.rs
```diff
@@ -368,6 +368,19 @@ macro_rules! store { (
...
+        pub fn activate(&self, ifs: &'static littlefs2::fs::Filesystem<$Ifs>,
+				efs: &'static littlefs2::fs::Filesystem<$Efs>,
+				vfs: &'static littlefs2::fs::Filesystem<$Vfs>) {
+            let store_ifs = $crate::store::Fs::new(ifs);
+            let store_efs = $crate::store::Fs::new(efs);
+            let store_vfs = $crate::store::Fs::new(vfs);
+            unsafe {
+		Self::ifs_ptr().write(store_ifs);
+		Self::efs_ptr().write(store_efs);
+		Self::vfs_ptr().write(store_vfs);
```
szszszsz, line 380: This looks like an invalid block indentation. Shall we introduce pre-commit checks?

daringer: fixed by replacing tabs with spaces, although this is actually something I would like `rust-fmt` to do in the future, currently running `rust-fmt --check` will lead to lots of changes - so I would suggest to leave this (and therefore `rust-fmt` enforcement) to be done soon(tm)

robin-nitrokey (commented):
The clients-n changes look good to me.  For the store change, I’ve added a comment regarding naming and consistency, otherwise looks good to me too.

PR #28, opened 2022-07-13
Add a `try_into_new_client(self, client_id: &str )` method to Service

sosthene-nitrokey:
This method is similar to `try_as_new_client` except the resulting ClientImplementation owns the Service. This makes it `'static` which is useful in many cases.

This required also implementing `Syscall` for `Service` (it was previously only implemented for `&mut Service`.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -751,3 +751,11 @@ where P: Platform
         self.process();
     }
 }
+
+impl<P> crate::client::Syscall for Service<P>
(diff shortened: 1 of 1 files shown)


robin-nitrokey (approved):

daringer:
- [x] does not change the existing API
- [x] github workflows run w/o issues
- [x] properly reviewed & tested

PR #29, opened 2022-07-17
[RFC] RSA lib integration

alt3r-3go:
This is an "request for comments"-type PR to invite a discussion and iron out the details of the RSA library integration.

#### Context
We've discussed this with the Nitrokey team a while ago (around the end of 2021) and I took a stab at it, then it got paused for several months after [this discussion](https://matrix.to/#/!zMQxxTSteoBoAeoObv:matrix.org/$kZ_aCPeh01bFwuARIJ1Q95nGeezJIVXbFhNtboNV6Iw?via=solokeys.com&via=matrix.org&via=nitro.chat) we additionally had with @nickray. Recently @daringer suggested we move ahead anyway, probably in a form of an NK-specific temporary fork, while those TIPs are getting written/implemented. AFAIU this is driven by the desire to have an RSA implementation for NK3, so at least some working solution is valued higher than the ideal one we discussed in that Matrix chat.

#### Implemented now
RSA2K-PKCS_v1.5 signing/verification and the whole necessary set of helper traits (`DeriveKey`, etc) + tests for them, that work on a PC. I separately was able to run Trussed and the RSA lib's full set of tests on my nRF52832, but haven't yet tried both of them at once.

#### Caveats
The PR is still well WIP and has items to discuss (marked as `TODO:alt3r-3go` in the code). I've based the general approach on the `miracl` one (#12) and that proved to be quite useful, so kudos to @nickray for such a nice example. Due to one of the TODOs related to the test infra, tests need to be run one by one for now, at the bottom of this description I'm showing a simple script that can be used for that. I haven't included it into the code, but let me know if it's helpful until I close the TODO.

All comments, big and small, as well as suggestions on the TODOs are more than welcome.

#### Testing script
```bash
#!/bin/bash

set -e
set -x

clear && RUST_BACKTRACE=1 cargo test --test rsa2kpkcs rsa2kpkcs_sign_verify -- --nocapture --show-output
RUST_BACKTRACE=1 cargo test --test rsa2kpkcs rsa2kpkcs_generate -- --nocapture --show-output
RUST_BACKTRACE=1 cargo test --test rsa2kpkcs rsa2kpkcs_derive -- --nocapture --show-output
RUST_BACKTRACE=1 cargo test --test rsa2kpkcs rsa2kpkcs_exists -- --nocapture --show-output
RUST_BACKTRACE=1 cargo test --test rsa2kpkcs rsa2kpkcs_serialize -- --nocapture --show-output
RUST_BACKTRACE=1 cargo test --test rsa2kpkcs rsa2kpkcs_deserialize -- --nocapture --show-output
RUST_BACKTRACE=1 cargo test --test rsa2kpkcs rsa2kpkcs_sign_verify -- --nocapture --show-output
```

### DIFF
--- a/src/mechanisms.rs
+++ b/src/mechanisms.rs
@@ -47,7 +47,8 @@ pub struct P256Prehashed {}
 mod p256;
 
 pub struct Rsa2kPkcs {}
-mod rsa2kpkcs;
+// Later on we'll add: "pub struct Rsa2kPss {}" and so on
+mod rsa2k;
 
 pub struct Sha256 {}
 mod sha256;
--- a/src/mechanisms/rsa2k.rs
+++ b/src/mechanisms/rsa2k.rs
@@ -12,11 +12,7 @@ use crate::error::Error;
 use crate::service::*;
 use crate::types::*;
 
-//TODO:alt3r-3go: sign() and verify() are the only two methods that are actually different between -pkcs and -pss.
-//                Moreover, the key::Kind::Rsa2K could also probably be parametrized, instead of having a dedicated kind
-//                for each. Overall this means the class structure can probably be simplified - need to decide.
-
-#[cfg(feature = "rsa2k-pkcs")]
+#[cfg(feature = "rsa2k")]
 impl DeriveKey for super::Rsa2kPkcs
 {
     #[inline(never)]
@@ -55,7 +51,7 @@ impl DeriveKey for super::Rsa2kPkcs
     }
 }
 
-#[cfg(feature = "rsa2k-pkcs")]
+#[cfg(feature = "rsa2k")]
 impl DeserializeKey for super::Rsa2kPkcs
 {
     #[inline(never)]
@@ -73,7 +69,7 @@ impl DeserializeKey for super::Rsa2kPkcs
         let private_key: RsaPrivateKey = DecodePrivateKey::from_pkcs8_der(&request.serialized_key)
             .map_err(|_| Error::InvalidSerializedKey)?;
 
-        // We store our keys in PKCS#8 DER format as well
+        // We store our keys in PKCS#8 DER format
         let private_key_der = private_key.to_pkcs8_der()
             .expect("Failed to serialize an RSA 2K private key to PKCS#8 DER");
 
@@ -88,7 +84,7 @@ impl DeserializeKey for super::Rsa2kPkcs
     }
 }
 
-#[cfg(feature = "rsa2k-pkcs")]
+#[cfg(feature = "rsa2k")]
 impl GenerateKey for super::Rsa2kPkcs
 {
     #[inline(never)]
@@ -125,7 +121,7 @@ impl GenerateKey for super::Rsa2kPkcs
     }
 }
 
-#[cfg(feature = "rsa2k-pkcs")]
+#[cfg(feature = "rsa2k")]
 impl SerializeKey for super::Rsa2kPkcs
 {
     #[inline(never)]
@@ -140,8 +136,6 @@ impl SerializeKey for super::Rsa2kPkcs
             .material;
 
         let serialized_key = match request.format {
-            // TODO:alt3r-3go: There are "Der" and "Asn1Der" commented out in KeySerialization enum,
-            //                 should those be used instead?
             KeySerialization::Raw => {
                 let mut serialized_key = Message::new();
                 serialized_key.extend_from_slice(&priv_key_der).map_err(|_| Error::InternalError)?;
@@ -155,7 +149,7 @@ impl SerializeKey for super::Rsa2kPkcs
     }
 }
 
-#[cfg(feature = "rsa2k-pkcs")]
+#[cfg(feature = "rsa2k")]
 impl Exists for super::Rsa2kPkcs
 {
     #[inline(never)]
@@ -169,7 +163,7 @@ impl Exists for super::Rsa2kPkcs
     }
 }
 
-#[cfg(feature = "rsa2k-pkcs")]
+#[cfg(feature = "rsa2k")]
 impl Sign for super::Rsa2kPkcs
 {
     #[inline(never)]
@@ -188,12 +182,9 @@ impl Sign for super::Rsa2kPkcs
             .expect("Failed to deserialize an RSA 2K private key from PKCS#8 DER");
 
         // RSA lib takes in a hash value to sign, not raw data.
-        // TODO:alt3r-3go: Do we assume we get digest into this function, or we calculate it ourselves?
-        // use sha2::digest::Digest;
-        // let digest_to_sign: [u8; 32] = sha2::Sha256::digest(&request.message).into();
+        // We assume we get digest into this function, too.
 
-        // TODO:alt3r-3go: There's also .sign_blinded(), which is supposed to protect the private key from timing side channels,
-        //                 but requires an RNG instance - decide if we want to always use it.
+        // TODO: Consider using .sign_blinded(), which is supposed to protect the private key from timing side channels
         use rsa::padding::PaddingScheme;
         use rsa::hash::Hash;
         let native_signature = priv_key
@@ -211,7 +202,7 @@ impl Sign for super::Rsa2kPkcs
     }
 }
 
-#[cfg(feature = "rsa2k-pkcs")]
+#[cfg(feature = "rsa2k")]
 impl Verify for super::Rsa2kPkcs
 {
     #[inline(never)]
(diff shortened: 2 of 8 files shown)


robin-nitrokey:
Thank you for the PR!  I did not have a closer look yet but I just want to give you a heads-up that we are already working on a solution for the test issue.

robin-nitrokey:
@alt3r-3go https://github.com/trussed-dev/trussed/pull/31 refactors the client used for the tests and should allow you to execute them normally with `cargo test --features virt`.

daringer:
Hey @alt3r-3go,

great work, I will answer (hopefully all) of your TODOs here to have it in one place.

* let's first stick (implementation-wise) to `rsa2k` only

* generally [cfg-feature granularity](https://github.com/trussed-dev/trussed/pull/29/files#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR48): I would suggest to have exactly 3 features for RSA: `rsa2k`, `rsa3k` and `rsa4k`

* don't worry about the [magic numbers](https://github.com/trussed-dev/trussed/pull/29/files#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR54) for now, `config.rs` anyways (as [this comment](https://github.com/trussed-dev/trussed/blob/3da56d8e41b9de64b852ed2bdcfd623118b09c3d/src/config.rs#L6) implies) lacks a proper magic-number-management-mechanism 

* using this configured numbers for [KeyStore](https://github.com/trussed-dev/trussed/pull/29/files#diff-42a890da1c9f63b32160f1b00e612dd86d9202febf0e8ccdc5543b6d47d9a25eR150) is perfectly fine for now - later this would need a compile-time constant (see magic-number-management-mechanism)

* essentially the same argument for [signature length](https://github.com/trussed-dev/trussed/pull/29/files#diff-f51aefff1ec1c07ff1b59d5f2546baaf3804af1c2cbd6c4881ce0251ed2a9435R226), funny though that `p256::Verify` does not check signature length at all and `ed255` does it based on a constant from `salty` - under the line: this is fine for now

* [serialization](https://github.com/trussed-dev/trussed/pull/29/files#diff-f51aefff1ec1c07ff1b59d5f2546baaf3804af1c2cbd6c4881ce0251ed2a9435R143) is perfectly fine with Raw only for now, keep it simple, as long as nothing forces us towards `der` or `asn1der` we stay with `raw`, which anyways implies pkcs#8-der (right?) 

* skip `sign_blinded` for now, to keep it simple (from [here](https://github.com/trussed-dev/trussed/pull/29/files#diff-f51aefff1ec1c07ff1b59d5f2546baaf3804af1c2cbd6c4881ce0251ed2a9435R195))

* [rsa takes a hash instead of raw data](https://github.com/trussed-dev/trussed/pull/29/files#diff-f51aefff1ec1c07ff1b59d5f2546baaf3804af1c2cbd6c4881ce0251ed2a9435R191), this makes it a "prehased" algortihm, thus we should stick to how `p256prehashed` does it: `request.message` directly contains the hash.

* sign & verify i.e. the [PSS vs. PKCS](https://github.com/trussed-dev/trussed/pull/29/files#diff-f51aefff1ec1c07ff1b59d5f2546baaf3804af1c2cbd6c4881ce0251ed2a9435R15) padding story: `p256` also implements two different [variants for sign](https://github.com/trussed-dev/trussed/blob/main/src/mechanisms/p256.rs#L236-L294), so RSA should do the same. This means we will have something like `rsa2k` as feature and then something like:
```
pub struct rsa2k {}
pub struct rsa2kpss {}
mod rsa2k;

// and then

impl Sign for super::rsa2k {}
impl Sign for super::rsa2kpss {}

impl Verify for super::rsa2k {}
impl Verify for super::rsa2kpss {}
```
But again: keep it simple and stick to `rsa2k` (pkcs) first...


Sooo, hope this answers the questions.

Rebasing on top of `main` also allowed me to run all the tests at once:
```
RUST_BACKTRACE=1 cargo test --test rsa2kpkcs --features virt -- --nocapture --show-output
```

This is really nice! Feels like nearly all of the major lifting stuff is done, despite the smaller changes described above (ok ignoring 3k, 4k - but that's totally fine for now). We anyways won't be able to upstream this until RSACrypto / RSA will have proper `no_std` support (which I cannot fully oversee what is missing at this point)

src/mechanisms.rs
```diff
@@ -47,7 +47,8 @@ pub struct P256Prehashed {}
 mod p256;
 
 pub struct Rsa2kPkcs {}
-mod rsa2kpkcs;
+// Later on we'll add: "pub struct Rsa2kPss {}" and so on
+mod rsa2k;
```
sosthene-nitrokey, line 51: Looks like you forgot to `git add` the new `src/mechanisms/rsa2.rs` after renaming.

szszszsz:
Hey @alt3r-3go !
Is there any chance you would have time to look into this PR again this week?

sosthene-nitrokey:
I mean't this PR should be reset to my https://github.com/sosthene-nitrokey/trussed/tree/rsa-lib-integration-fmt branch to resolve merge conflicts

szszszsz:
Hey @sosthene-nitrokey @alt3r-3go ! 
Just FYI, I've added public key serialization and decryption support below. How to proceed to integrate? I can make a PR to this PR in a few days (feel free to do it earlier if you need it):
- https://github.com/trussed-dev/trussed/compare/6f4b547fe710058033efbb9090bcf5649c7a0494...Nitrokey:trussed:webcrypt-devel-rsa

sosthene-nitrokey:
I'm aiso working on a modified branch on top of #51: [here](https://github.com/sosthene-nitrokey/trussed/tree/rsa-rebased)

src/mechanisms/rsa2k.rs
```diff
@@ -0,0 +1,251 @@
...
+        let priv_key: RsaPrivateKey = DecodePrivateKey::from_pkcs8_der(&priv_key_der)
+            .expect("Failed to deserialize an RSA 2K private key from PKCS#8 DER");
+
+        // RSA lib takes in a hash value to sign, not raw data.
+        // We assume we get digest into this function, too.
+
+        // TODO: Consider using .sign_blinded(), which is supposed to protect the private key from timing side channels
+        use rsa::padding::PaddingScheme;
+        use rsa::hash::Hash;
+        let native_signature = priv_key
```
sosthene-nitrokey, line 190: This can panic il the length of the data is not `32`, it should return an error instead.

szszszsz:
@sosthene-nitrokey @daringer @robin-nitrokey  We should discuss this tomorrow

src/mechanisms/rsa2k.rs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        // TODO: This must not be a hardcoded magic number, convert when a common mechanism is available
+        if request.signature.len() != 256 {
+            return Err(Error::WrongSignatureLength);
+        }
+
+        let key_id = request.key;
+
+        let priv_key_der = keystore.load_key(key::Secrecy::Secret, Some(key::Kind::Rsa2k), &key_id)
```
sosthene-nitrokey, line 224: Signature verification should probably load a public key.

src/mechanisms/rsa2k.rs
sosthene-nitrokey: Also we may want to use another hash than SHA-256. I'd be in favor of using `None` for the hash. We need that for OpenPGP anyway, as all hashes can be used. it depends on the host.

It would require added documentation that the hash prefix needs to be added by the client though.

sosthene-nitrokey: I agree this is somewhat uncomfortable, but for [OpenPGP](https://raw.githubusercontent.com/Nitrokey/openpgp-card/master/OpenPGP%20Card%20Specification%203.4.pdf): in § 7.2.10, it mentions the digest prefix added for each hash that come as part of the input (DigestInfo) "The card may not check the correctness of the DigestInfo.". So we need to accept any input and not attempt to parse it or add any prefix other than the Pkcs#1v15 padding.

src/mechanisms/rsa2k.rs
sosthene-nitrokey: The keystore can hold both public and private keys. It's a bit of a shame that it needs to be told what key is being loaded rather than returning an enum allowing to match against the key type, which would be more flexible.

sosthene-nitrokey: > It does derive it, a little below.

Yes, but the current implementation means you can't verify a signature unless you also have the private key.

PR #30, opened 2022-07-20
Add metadata method to the FilesystemClient and Filestore traits

sosthene-nitrokey:
This method allows efficient probing of files and directories, and is especially useful to know if they exist or not.


robin-nitrokey (approved):

daringer:
- [ ] does not change the existing API
- [ ] github workflows run w/o issues
- [ ] properly reviewed

PR #31, opened 2022-07-20
Add virtual platform

robin-nitrokey:
This patch adds a `virt` feature and module with a virtual platform implementation.  This platform uses a basic user interface implementation and RAM storage.  It makes it easy to run tests without having to implement all required Trussed components.  As the `TrussedInterchange` and the RAM storage cannot be used concurrently, a mutex is used to protect the platform.

This is based on the platform that was already used in Trussed’s tests (`tests/card/mod.rs`).

This PR depends on #28 because of lifetime issues.


sosthene-nitrokey:
```
error[E0282]: type annotations needed
  --> tests/counter.rs:36:18
   |
36 |     client::get(|client| {
   |                  ^^^^^^ consider giving this closure parameter a type
   |
   = note: type must be known at this point
```

Tests don't  compile without `--features virt`. They should probably be gated behind  a feature flag.

sosthene-nitrokey:
I think optional features are not compiled on docs.rs, do we want to add
```toml
[package.metadata.docs.rs]

features = ["virt"]
```

 in Cargo.toml so that the virtual client is documented on docs.rs?

robin-nitrokey:
Thanks for the review!  I’ve added a `cfg` attribute to the tests and activated the `virt` feature for `docs.rs`.  Ideally, we would have `virt` as a required feature for the tests in the manifest, but that would require a separate entry for each file which is not really nice to maintain.

daringer:
uh great for testing!

PR #34, opened 2022-07-22
Add a entry_metadata method to FilesystemClient

sosthene-nitrokey:
This should have been part of #30 but I forgot to push it :sweat_smile: 

This adds some basic filesystem tests as well.


daringer:
let's go

PR #39, opened 2022-08-19
Add support for extracting keys

sosthene-nitrokey:
This PR follows what was discussed in https://github.com/trussed-dev/trussed/discussions/36.

There are two issues to solve:

- [ ] Migrations: Already created public keys will have the `EXTRACTABLE` flag off by default, when it should be set, otherwise it will not be possible to serialize them.
- [ ] What mechanism should be used to serialize a symmetric key?

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -79,6 +79,10 @@ bitflags::bitflags! {
         const LOCAL = 1 << 0;
         /// Set if the key is a secret key.
         const SENSITIVE = 1 << 1;
+        // Reserved for future use
+        // const WRAPPABLE = 1 << 3;
+        /// This flag currently only applies to `kind::Shared`
+        const SERIALIZABLE = 1 << 4;
     }
 }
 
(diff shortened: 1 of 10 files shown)


sosthene-nitrokey:
This PR now adds 
- a `SERIALIZABLE` key flag
- A Serializable `StorageAttribute` which is only considered in `CryptoClient::agree` to set the `SERIALIZABLE` key flag
- A `SharedSecret` `Mechanism` used to serialize `kind::Shared` with the `SERIALIZABLE` flag set.

sosthene-nitrokey:
This PR now adds 
- a `SERIALIZABLE` key flag
- A Serializable `StorageAttribute` which is only considered in `CryptoClient::agree` to set the `SERIALIZABLE` key flag
- A `SharedSecret` `Mechanism` used to serialize `kind::Shared` with the `SERIALIZABLE` flag set.

sosthene-nitrokey:
The build failure seems to be caused by https://github.com/trussed-dev/interchange/pull/4, and that the solo2 firmware uses undocumented APIs that were modified.

robin-nitrokey:
Will be fixed by https://github.com/solokeys/solo2/pull/122.

nickray:
@sosthene-nitrokey can you please rebase on main (fixing fake merge conflicts), then feel free to merge.

PR #41, opened 2022-09-05
Add multiple (Service) Backends abstraction

daringer:
This adds the structure to allow multiple service backends. Essentially the client now has the possibility to choose a order-of-dispatch in its `ClientContext`. This is an elementary step towards transparent hardware crypto devices like the se050.

* Extend `ClientId` to be a proper struct and call it make it a `ClientContext` (as the name suggests)
* Fix several concurrency issues during directory traversal using the `ClientContext` instead of a quasi-global
* extend all necessary components and `reply_to` dispatching based on client-chosen service backends

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -241,30 +241,57 @@ the walker state of the directory traversal syscalls.
 */
 pub struct ClientContext {
     pub path: PathBuf,
+    pub backends: Vec<ServiceBackends, 2>,
     pub(crate) read_dir_state: Option<ReadDirState>,
     pub(crate) read_dir_files_state: Option<ReadDirFilesState>,
 }
 
 impl core::convert::From<PathBuf> for ClientContext {
     fn from(path: PathBuf) -> Self {
-        Self::new(path)
+        Self::from_pathbuf(path)
     }
 }
 
 impl core::convert::From<&str> for ClientContext {
     fn from(path_str: &str) -> Self {
-        Self::new(PathBuf::from(path_str))
+        Self::from_pathbuf(PathBuf::from(path_str))
     }
 }
 
 impl ClientContext {
-    pub fn new(path: PathBuf) -> Self {
+    pub fn from_pathbuf(path: PathBuf) -> Self {
+        Self::new(path, Vec::from_slice(&[ServiceBackends::Software]).unwrap())
+    }
+
+    pub fn new(path: PathBuf, backends: Vec<ServiceBackends, 2>) -> Self {
         Self {
             path,
+            backends,
             read_dir_state: None,
             read_dir_files_state: None,
         }
     }
+}
+
+/**
+This enum collects the available "system call" service backends.
+Each of these backends is supposed to be self-sufficient (i.e. not reliant
+on another one or on ServiceResources itself) and to implement a subset
+of the API calls.
+
+If a backend requires ownership of a device, that device will be part of
+the customizable part of the macro-constructed Platform struct. If a
+backend further requires parametrization from the client to operate (such as
+a PIN being passed down from an app), this parameter struct should be
+added as payload for that enum variant.
+
+Backends are called from Service::process() under consideration of the
+selection and ordering the calling client has specified in its ClientId.
+*/
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum ServiceBackends {
+    Software,
+    // SE050(Se050Parameters),
 }
 
 // Object Hierarchy according to Cryptoki
--- a/src/types.rs
+++ b/src/types.rs
@@ -19,6 +19,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::config::*;
 use crate::key::Secrecy;
+use crate::store::filestore::{ReadDirFilesState, ReadDirState};
 
 pub use crate::client::FutureResult;
 pub use crate::platform::Platform;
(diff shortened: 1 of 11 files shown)


sosthene-nitrokey:
I've rebased it on top of 8e347abf99dba81fca11c5779a473d263a7b0565 in [sosthene-nitrokey:client-context-rebased](https://github.com/sosthene-nitrokey/trussed/tree/client-context-rebased)

src/types.rs
```diff
@@ -72,11 +75,6 @@ impl Id {
         let array = self.0.to_be_bytes();
 
         for i in 0..array.len() {
-            if array[i] == 0 && i != (array.len() - 1) {
```
sosthene-nitrokey, line 75: If leading zeros are not skipped anymore this comment should probably be removed

daringer: good point, fixed

nickray: I fear this will break backwards compatibility, no?

src/types.rs
```diff
@@ -232,8 +229,105 @@ pub mod consent {
 // pub type AeadNonce = [u8; 12];
 // pub type AeadTag = [u8; 16];
 
-// pub type ClientId = heapless::Vec<u8, heapless::consts::U32>;
-pub type ClientId = PathBuf;
+/**
+The "ClientId" struct is the closest equivalent to a PCB that Trussed
```
nickray, line 233: What is a PCB?

daringer: Process Control Block, OS "slang". If Trussed is an OS, the ClientContext struct would be roughly an equivalent to a PCB for clients ('clients' could then be understood as processes)

robin-nitrokey:
I think we could further modularize this PR by moving the backends out of Trussed.  The backend enum could be defined together with the platform by the runner.  The backend implementations (other than the software backend) could reside in separate crates.  This would mean that we would have to make `Request` and the types that use it generic over the additional backends, but I think it would be a cleaner approach and would cause less overhead.

I’m currently polishing a draft implementation and will share it here in the next days for you to review.

nickray:
Agreed that any and all non-default backends should be independent crates, and the integration designed as such.
Honestly, even the software backend shouldn't be hardcoded into Trussed, but likely generated based on "constraints" of what is needed and which low-level implementations to use.

The generic approach has a danger of leading to a deluge of generics; this could be improved with more code generation (essentially the hardcoded/upstreamed approach, but on-demand and dynamic).

robin-nitrokey:
For reference, here is my current implementation:
- https://github.com/Nitrokey/trussed/pull/9
- https://github.com/Nitrokey/trussed/pull/10

I’ll update this PR once we’ve finished testing the integration into the firmware.  While this adds some additional type parameters, I think the approach using the const-generics interchange implementation is a good trade-off between type complexity and flexibility.

robin-nitrokey:
>  this could be improved with more code generation 

I’d rather reduce the amount of code generation in Trussed because I feel that the code generation macros often are not flexible enough and create additional restrictions (e. g. `platform!` does not support generics).  My suggestion would be to provide simple defaults for the trivial cases (e. g. no custom backends) while allowing users to manually implement more complex cases.

nickray:
Specific to your alternative PR, we had a lot more generic parameters in the past and got rid of them. E.g., code like
```
impl<B: 'static, I: TrussedInterchange<B>, S: Syscall> HmacBlake2s
    for ClientImplementation<B, I, S>
```
is repetitive and quite annoying to maintain. It will just get worse once we wish to add more flexibility. It also has no actual value in flexibility, as things are set (I think?) once and for all in setup of an actual build.

Understood that the current "macros by example" are limited. I'm not thinking of even more such macros, but rather a formal step that could go two ways (that I can think of):

a) in a build.rs that "somehow" is configured (it isn't really idiomatic to do a lot of work in build.rs though)
a') annoying alternative pre-build step, that is more flexible than build.rs
b) the "config crate idea", which I first heard of from James Munns

What is the config crate? Say we have a type like
```
    if #[cfg(feature = "clients-12")] {
        interchange::interchange! {
            TrussedInterchange: (Request, Result<Reply, Error>, 12)
        }
```
Here the `clients-*` features are already a hack, as Cargo features are only booleans.

The idea is this: Let `trussed` depend on `trussed-config`, which has some kind of default type definitions (e.g. TrussedInterchange as an interchange for 3 clients). Then inside `trussed`, we somewhere have
```
pub mod config {
  pub use trussed_config::*;
}
```
and everything else depends on `trussed::config`.

In firmware code (or PC runners, or examples, etc.), one defines a local crate (defining e.g. TrussedInterchange with 12 clients), and has a `[patch]` section in the main Cargo.toml to use this instead.

This approach localizes all the things that must vary in one place; the code depending on it in `trussed` has no generics anymore.

As a first step, these config crates would be "written by hand", but that gets painful too.

So then ideally, we have an actual code generator *just for this* (i.e. not Rust macros, but an actual dedicated software in Rust or Python), which itself can be driven from a TOML or other config *file*.

Or maybe even the users don't need to supply these crates, and just supply a trussed.toml config file, which `trussed`'s `build.rs` uses to run the code generator on.

An alternative is to tie all these generic parameters here, so we'd still model them properly in your sense, but `trussed` only sees the generic-free specific instantiation.

This build step could/should also be able to run the "dependency resolver" that analyzes apps for their cryptographic and other needs, and generates a specific software backend that has just that.

Understood that this is just a sketch and at most a basis for a deeper discussion. I do think that the question "how do we configure Trussed" is core and central to Trussed's success, and applicability in a variety of use cases (even just SoloKeys vs Nitrokey, but even more so beyond these to desktops + servers).

nickray:
Yes, a Config trait might work for some use cases too. The overarching message I'm trying to send is to keep the overall generic count in any given part of the code as low as possible.

That's a cute trick to count the clients :)

robin-nitrokey:
Closing in favor of https://github.com/trussed-dev/trussed/pull/68

PR #43, opened 2022-09-13
Inject any key

szszszsz:
Draft solution for injecting any key material with the specified kind to the Trussed system.
Such is required for Nitrokey Webcrypt and OpenPGP implementations.

Requesting for comments.
To be rebased on the current master.


sosthene-nitrokey:
It seems to me like reviving `unsafe_inject_key` would be the proper method, I don't really know why it was deprecated. At the time there was no `SharedSecret` mechanism, so maybe that's why it was changed? @nickray can you confirm? 915c544c94049e60eb8d82f250450f5e9f3c80a8 doesn't tell much more.

> add "imported" attribute to the key, allowing to identify it when used with security critical functions

Isn't it redundant with the `LOCAL` flag?

src/client.rs
```diff
@@ -544,10 +546,12 @@ pub trait CryptoClient: PollClient {
         &mut self,
         raw_key: &[u8],
         location: Location,
+        kind: Kind,
```
sosthene-nitrokey, line 549: `key::Kind` is not exposed in the `Client` API currently, I'm not sure it should. To me it's more an internal detail an the `mechanism` enum should be used instead.

szszszsz:
> > add "imported" attribute to the key, allowing to identify it when used with security critical functions
> Isn't it redundant with the `LOCAL` flag?

It is indeed, did not notice that before. I would suggest changing the name to `LOCAL_ORIGIN`, as this sounds confusingly similar to resident/stored keys.

src/client.rs
szszszsz: If so, then surely we can go to the original API, without the needless extension.

sosthene-nitrokey:
This is turning out to be much more complex than expected for curve25519 keys. It is fine for `P256` though.

Currently we use [`salty`](https://docs.rs/salty/latest/salty/) for curve25519. The issue is that it doesn't have any way to construct a private key from its raw byte representation. Keys are stored by their "seed" which is then hashed to obtain the private key. This scheme cannot be used to store imported keys as it would require reversing the hash.

So first we'll need a way to create `salty::SecretKey` from their raw representation. Then we'll need to either migrate the key representation to a new one, or simply ignore the old keys if no current deployement of `trussed` actually uses X255 keys.

sosthene-nitrokey:
A solution could be to handle it by using either the `kind` enum for the key or a flag maybe even the `LOCAL` flag to differentiate keys stored as their seed and keys stored as their value.

sosthene-nitrokey:
At least it lead to https://github.com/ycrypto/salty/pull/28

sosthene-nitrokey:
@szszszsz You can remove the draft status on this PR, it works correctly for opcard.

PR #44, opened 2022-09-13
Add P256 HMAC-based key derivation from the supplied data

szszszsz:
Based on the HmacSha256 implementation - the only difference is, that the result is stored as a P256 key, instead of a shared variant.

As discussed, it would be best in the future to introduce shared implementation across different combinations of hash and resulting key algorithms.

Requesting for comments / solution suggestions.

-----

Improvement ideas:
- Provide a way to connect multiple mechanisms together, without copying the implementation like in this case.

Current downsides:
- supplied implementation is almost a verbatim copy of `HmacSha256`.

-----

To do:
- [ ] change the algorithm to seed-based Ed25519 instead
- [ ] make sure the generated keys are correct - apply proper checks

-----

Note: this PR will be soon obsolete, in a favor of using backend and extension based implementation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,7 @@ default-mechanisms = [
     # "hmac-blake2s",
     "hmac-sha1",
     "hmac-sha256",
+    "hmac-sha256-p256",
 	# For some reason, this breaks Solo 2 firmware
 	# At minimum, this seems to have a huge "block" method
     # "hmac-sha512",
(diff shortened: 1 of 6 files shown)


szszszsz:
Yes, switch to Ed25519 makes sense. Will do so in the next step, and remove P256 usage completely.

robin-nitrokey:
Closing as this has been implemented as a custom backend for websmartcard, see: https://github.com/Nitrokey/nitrokey-websmartcard-rust/pull/16

PR #51, opened 2022-10-06
Upgrade dependencies

sosthene-nitrokey:


sosthene-nitrokey:
This PR updates all dependencies. This is required to use recent versions of Zeroize because chacha20 v0.7 has an upper bound to 1.4.

sosthene-nitrokey:
We need to check that the postcard update does not break anything

robin-nitrokey (approved):
We already have this in https://github.com/Nitrokey/trussed so it should be good to merge.

robin-nitrokey:
Please rebase and drop the postcard update and revert commits.

sosthene-nitrokey:
I rebased it.

robin-nitrokey:
Thanks!  When updating our fork, I realized that updating blake2 breaks the HmacBlake2s mechanism.  This will probably cause the CI to fail.  It should be easy to fix by replacing `Hmac<blake2::Blake2s>` with `SimpleHmac<blake2::Blake2s256>` and removing the unused `NewMac` import.

sosthene-nitrokey:
I rebased it on top of latest main. 
This upgrades all dependencies that can be upgraded without issues.

- Bitflags will be upgraded in a separate PR (looks like it's much less straightforward than the other ones.
- Postcard is not updated to not break compatibility with deployed firmware.

sosthene-nitrokey:
I rebased it on top of latest main. 
This upgrades all dependencies that can be upgraded without issues.

- Bitflags will be upgraded in a separate PR (looks like it's much less straightforward than the other ones.
- Postcard is not updated to not break compatibility with deployed firmware.

PR #58, opened 2022-12-01
Relax key kind check in derive_key for hmacsha256

robin-nitrokey:
In #38 [0], we restricted the operations for the hmacsha256 mechanism to symmetric and shared keys.  This breaks fido-authenticator because it uses hmacsha256 derive_key on EC keys [1].  This patch relaxes the restrictions to allow all key kinds.

[0] https://github.com/trussed-dev/trussed/pull/38
[1] https://github.com/solokeys/fido-authenticator/issues/21

cc @sosthene-nitrokey: I can’t add you as a reviewer, but please have a look.

### DIFF
--- a/src/mechanisms/hmacsha256.rs
+++ b/src/mechanisms/hmacsha256.rs
@@ -16,7 +16,9 @@ impl DeriveKey for super::HmacSha256 {
         let key_id = request.base_key;
         let key = keystore.load_key(key::Secrecy::Secret, None, &key_id)?;
         if !matches!(key.kind, key::Kind::Symmetric(..) | key::Kind::Shared(..)) {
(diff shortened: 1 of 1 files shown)


sosthene-nitrokey:
We should probably keep the check and add a `warn!` log so that we can see when it is relied upon and to make it less likely that we rely upon it by mistake in new updates.

robin-nitrokey:
Good point, done.

sosthene-nitrokey (approved):

daringer (approved):

nickray:
Philosophically, I think we could allow (or at least imagine) casting an asymmetric secret key to a "shared secret", which is always only used as input to a KDF kind of function.

PR #60, opened 2022-12-15
virt: Improve storage implementation

robin-nitrokey:
This PR contains multiple improvements to the virtual platform and the virtual storage implementations.  It should make the platform more stable and contains all features required for usbip simulation of a complete device including the provisioner app.


szszszsz (approved):
Looks good!

PR #61, opened 2022-12-15
Build embedded runner for nk3 in CI

robin-nitrokey:


daringer (approved):

nickray:
Maybe we can parallelise these firmware builds.

Issue #65, opened 2023-01-17
Restrict file paths for FilesystemClient syscalls


robin-nitrokey:
All functions in `FilesystemClient` should make sure that the client does not escape its `dat` namespace.

https://github.com/trussed-dev/trussed/blob/9b9662d08ec4b8d653d1ba8a3aa486681cea8b33/src/store/filestore.rs#L43-L49

PR #66, opened 2023-01-19
Enforce namespace in FilesystemClient

robin-nitrokey:
This patch adds a check to FilesystemClient that makes sure that clients cannot read out files that belong to other clients or that are not in the data directory for the client.

This is a breaking change as the Error enum is extended.

Fixes https://github.com/trussed-dev/trussed/issues/65

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed off-by-one error in `RandomBytes` request.
 - Fixed a race condition when iterating over the filesystem in more than one
   client ([#64]).
+- Fixed missing path validation in `Filestore` that allowed clients to escape
+  their namespace ([#65]).
 
 [#64]: https://github.com/trussed-dev/trussed/issues/64
+[#65]: https://github.com/trussed-dev/trussed/issues/65
 
 ## [0.1.0] - 2022-01-26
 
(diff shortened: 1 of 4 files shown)


src/mechanisms/hmacsha256.rs
```diff
@@ -18,7 +18,10 @@ impl DeriveKey for super::HmacSha256 {
         if !matches!(key.kind, key::Kind::Symmetric(..) | key::Kind::Shared(..)) {
             // We have to disable this check for compatibility with fido-authenticator, see:
             // https://github.com/solokeys/fido-authenticator/issues/21
-            warn!("derive_key for hmacsha256 called with invalid key kind ({:?})", key.kind);
+            warn!(
+                "derive_key for hmacsha256 called with invalid key kind ({:?})",
+                key.kind
+            );
```
szszszsz, line 24: This change looks unrelated

robin-nitrokey: moved to: https://github.com/trussed-dev/trussed/pull/69

szszszsz: It's still here, unless there is something with my GH view.

robin-nitrokey: Yes, this PR has not been rebased yet.  But if you compare to current main, there are no changes to `src/mechanisms/hmacsha256.rs` anymore.

szszszsz: Right. Resolving.

robin-nitrokey:
Rebased onto 88de4c6b3ac8c94534aad98efbed02af8eaec70c.

robin-nitrokey:
Rebased and changed to just forbid `..` for performance reasons.

PR #68, opened 2023-01-20
Add support for multiple backends

robin-nitrokey:
This is a reduced version of https://github.com/trussed-dev/trussed/pull/41 as discussed today.  The main difference is that the service endpoints cannot be configured at runtime (using a syscall), but only during client creation.  This makes the interface much simpler.  I did not convert the backend ID to an integer as originally planned because there is only one remaining struct that uses the backend ID, `ServiceEndpoint`.

Based on https://github.com/trussed-dev/trussed/pull/67

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -11,7 +11,7 @@ use crate::{
     error::Error,
     platform::Platform,
     service::ServiceResources,
-    types::{ClientContext, Empty},
+    types::{Context, CoreContext, Empty, NoData},
 };
 
 /// The ID of a backend.
@@ -25,14 +25,21 @@ pub enum BackendId<I> {
 
 /// A custom backend that can override the core request implementations.
 pub trait Backend<P: Platform> {
+    /// The context for requests handled by this backend.
+    type Context: Default;
+
     /// Executes a request using this backend or returns [`Error::RequestNotAvailable`][] if it is
     /// not supported by this backend.
     fn request(
         &mut self,
-        client_ctx: &mut ClientContext,
+        core_ctx: &mut CoreContext,
+        backend_ctx: &mut Self::Context,
         request: &Request,
         resources: &mut ServiceResources<P>,
-    ) -> Result<Reply, Error>;
+    ) -> Result<Reply, Error> {
+        let _ = (core_ctx, backend_ctx, request, resources);
+        Err(Error::RequestNotAvailable)
+    }
 }
 
 /// Dispatches requests to custom backends.
@@ -44,13 +51,15 @@ pub trait Backend<P: Platform> {
 pub trait Dispatch<P: Platform> {
     /// The ID type for the custom backends used by this dispatch implementation.
     type BackendId: 'static;
+    /// The context type used by this dispatch.
+    type Context: Default;
 
     /// Executes a request using a backend or returns [`Error::RequestNotAvailable`][] if it is not
     /// supported by the backend.
     fn request(
         &mut self,
         backend: &Self::BackendId,
-        ctx: &mut ClientContext,
+        ctx: &mut Context<Self::Context>,
         request: &Request,
         resources: &mut ServiceResources<P>,
     ) -> Result<Reply, Error> {
@@ -65,4 +74,5 @@ pub struct CoreOnly;
 
 impl<P: Platform> Dispatch<P> for CoreOnly {
     type BackendId = Empty;
+    type Context = NoData;
 }
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -1,0 +1,68 @@
+//! Custom backends that can override core request implementations.
+//!
+//! Trussed provides a default implementation for all [`Request`][]s, the core backend.  Runners
+//! can add custom [`Backend`][] implementations using the [`Dispatch`][] trait that can override
+//! the implementation of one or more requests.  The backends used to execute a request can be
+//! selected per client when constructing a client using
+//! [`ClientBuilder::backends`][`crate::client::ClientBuilder::backends`].
+
+use crate::{
+    api::{Reply, Request},
+    error::Error,
+    platform::Platform,
+    service::ServiceResources,
+    types::{ClientContext, Empty},
+};
+
+/// The ID of a backend.
+///
+/// This ID can refer to the core backend provided by the Trussed crate, or to a custom backend
+/// defined by the runner.  The custom ID type is defined by [`Dispatch::BackendId`][].
+pub enum BackendId<I> {
+    Core,
+    Custom(I),
+}
+
+/// A custom backend that can override the core request implementations.
+pub trait Backend<P: Platform> {
+    /// Executes a request using this backend or returns [`Error::RequestNotAvailable`][] if it is
+    /// not supported by this backend.
+    fn request(
+        &mut self,
+        client_ctx: &mut ClientContext,
+        request: &Request,
+        resources: &mut ServiceResources<P>,
+    ) -> Result<Reply, Error>;
+}
+
+/// Dispatches requests to custom backends.
+///
+/// If a runner does not support custom backends, it can use the [`CoreOnly`][] dispatch.
+/// Otherwise it can provide an implementation of this trait that defines which backends are
+/// supported.  The backends that are used to execute a request can be selected when constructing a
+/// client using [`ClientBuilder::backends`][`crate::client::ClientBuilder::backends`].
+pub trait Dispatch<P: Platform> {
+    /// The ID type for the custom backends used by this dispatch implementation.
+    type BackendId: 'static;
+
+    /// Executes a request using a backend or returns [`Error::RequestNotAvailable`][] if it is not
+    /// supported by the backend.
+    fn request(
+        &mut self,
+        backend: &Self::BackendId,
+        ctx: &mut ClientContext,
+        request: &Request,
+        resources: &mut ServiceResources<P>,
+    ) -> Result<Reply, Error> {
+        let _ = (backend, ctx, request, resources);
+        Err(Error::RequestNotAvailable)
+    }
+}
+
+/// Always dispatches to the Trussed core backend.
+#[derive(Debug, Default)]
+pub struct CoreOnly;
+
+impl<P: Platform> Dispatch<P> for CoreOnly {
+    type BackendId = Empty;
+}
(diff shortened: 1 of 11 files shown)


robin-nitrokey:
This backends implementation only changes the public interface for a single function, `Service::add_endpoint`, which is probably not used outside Trussed anyway.  If this is a concern, we can introduce a variant to maintain compatibility.

robin-nitrokey:
Rebased onto 88de4c6b3ac8c94534aad98efbed02af8eaec70c.

nickray (changes requested):
This can be reverted, https://github.com/solokeys/solo2/pull/152 is merged. Thanks for that!

nickray (changes requested):
I explored the suggestions concretely here: https://github.com/robin-nitrokey/trussed/compare/backends...nickray:trussed:elaboration

If this seems OK to you, happy to push to your branch and merge this PR.

robin-nitrokey:
Thanks for the review!  While working on https://github.com/trussed-dev/trussed/pull/70 I realized that the current `Backends::select` method is not sufficient:  We have to add an associated type for the context to `Backend` which makes the trait non-object-safe, so we cannot return a reference to it.  Instead, we directly have to call the `Backend::reply_to` method for a given backend:

```rust
pub trait Backends<P: Platform> {
    type Id: 'static;
    type ExtensionId: TryFrom<u8, Error = Error>;
    type Context: Default;

    fn core_request(
        &mut self,
        id: &Self::Id,
        ctx: &mut Context<Self::Context>,
        request: &Request,
        resources: &mut ServiceResources<P>,
    ) -> Result<Reply, Error> {
        let _ = (id, ctx, request, resources);
        Err(Error::RequestNotAvailable)
    }

    fn extension_request(
        &mut self,
        backend: &Self::Id,
        extension: &Self::ExtensionId,
        ctx: &mut Context<Self::Context>,
        request: &request::Extension,
        resources: &mut ServiceResources<P>,
    ) -> Result<reply::Extension, Error> {
        let _ = (backend, extension, ctx, request, resources);
        Err(Error::RequestNotAvailable)
    }
}
```

So I would rather keep the `Backends` name or maybe use something more generic like `Dispatch`.

nickray:
Dispatch sounds good. I find Backends too close to Backend. From the sound of it, though this PR needs to simmer a bit until #70 is more final?

robin-nitrokey:
Okay, I’ll change it to `Dispatch`.  #70 would already be ready to review if this is merged.  I split it up into two PRs to make it easier to review, but if you’d prefer doing both in one PR, we can also do that.

robin-nitrokey:
I’ve updated the PR:
- Rebased onto 1f3cb9db10c21a57f871725b339677c936a61b47
- Renamed `Backends` to `Dispatch` per our discussion
- Added a commit that introduces backend-specific context (extracted from #70)

daringer (approved):

src/client.rs
```diff
@@ -685,6 +687,70 @@ pub trait UiClient: PollClient {
...
+    pub fn new(id: impl Into<PathBuf>) -> Self {
+        Self {
+            id: id.into(),
+            backends: &[],
+        }
+    }
+}
+
+impl<B: 'static> ClientBuilder<B> {
+    pub fn with_backends<C: 'static>(self, backends: &'static [BackendId<C>]) -> ClientBuilder<C> {
```
sosthene-nitrokey, line 705: I would remove the `with`, it is generally not used in the builder pattern.

robin-nitrokey:
Updated:
- Rebased onto 47afe55313ef418c9448b690c209a3c4f1c8fd64.
- Renamed `ClientContext` to `CoreContext`.
- Renamed `with_backends` in `ClientBuilder` to `backends`.

src/backend.rs
```diff
@@ -0,0 +1,48 @@
...
+        ctx: &mut Context<Self::Context>,
+        request: &Request,
+        resources: &mut ServiceResources<P>,
+    ) -> Result<Reply, Error> {
+        let _ = (backend, ctx, request, resources);
+        Err(Error::RequestNotAvailable)
+    }
+}
+
+impl<P: Platform> Dispatch<P> for () {
```
nickray, line 45: I would prefer having this implemented on a specific type `struct SoftwareOnly` instead of using the unit type (name to be bikeshed). Think this is more clear.

src/backend.rs
```diff
@@ -0,0 +1,48 @@
...
+    type BackendId = ();
```
nickray, line 46: I would prefer using a void type `enum Empty {}` here instead of using the unit type (name to be bikeshed).
Cf. https://github.com/robin-nitrokey/trussed/compare/backends...nickray:trussed:elaboration even if the Dispatch method changed now. This enforces that you can't even generate a BackendId::Custom.

src/backend.rs
```diff
@@ -0,0 +1,48 @@
...
+    type Context = ();
```
nickray, line 47: Not sure what a "more correct" default type (empty struct vs empty enum) and name would be here.

src/client.rs
```diff
@@ -685,6 +687,69 @@ pub trait UiClient: PollClient {
...
+            .map_err(|_service_endpoint| ())?;
+        Ok(requester)
+    }
+
+    #[allow(clippy::result_unit_err)]
+    pub fn build<P: Platform, D: Dispatch<P, BackendId = B>, S: Syscall>(
+        self,
+        service: &mut Service<P, D>,
+        syscall: S,
+    ) -> Result<ClientImplementation<S>, ()> {
```
nickray, line 729: In these Results with trivial error types, maybe we should use Option and mark with `#[must_use]`?
I'm finding the proliferation of `()` a bit hard to follow.
Or would this break `?` in certain places?

nickray (changes requested):

src/client.rs
robin-nitrokey: I just used `()` for consistency with the existing methods, but I agree that `Error` is much better.  I’ll add a `ClientCountExceeded` variant and change all relevant methods in `ClientBuilder` and `Service` to return that error instead.

src/backend.rs
robin-nitrokey: I think the empty tuple is fine here because it just means “no data”.

src/backend.rs
robin-nitrokey: Done.

robin-nitrokey: Could probably be `!` once it’s stable.

robin-nitrokey:
Updated:
- Added `Error::ClientCountExceeded` variant and used it in the relevant methods.
- Added `SoftwareOnly` struct as the baseline `Dispatch` implementation.
- Added `Empty` enum.

I’m not sure whether “Software” is the best term to describe the Trussed default backend.  I think it was chosen with backends like SE050 in mind, but we will probably also have software-only backends that just implement extensions.  Maybe “Core” would be better?

src/backend.rs
nickray: So `NoData`? :)

src/client.rs
nickray: Yeah that works. Generally it'd be better to have tight error types (i.e. a client of just this method would know which error variants it can actually expect), but doing this in Rust is tedious as you'd need tons of conversion traits.

nickray:
I suggested `SoftwareOnly` for consistency with `BackendId::Software` (so change both if you change one). `Core` would be consistent with your `CoreContext`. `Default` or `Baseline` or whatever alternatives... Agree that the configured default backend doesn't necessarily have to be software, nor a custom backend non-software.

src/backend.rs
```diff
@@ -0,0 +1,43 @@
+use crate::{
+    api::{Reply, Request},
+    error::Error,
+    platform::Platform,
+    service::ServiceResources,
+    types::{ClientContext, Empty},
+};
+
+pub enum BackendId<C> {
```
nickray, line 9: Can we have `I` instead of `C` here please?

src/client.rs
```diff
@@ -685,6 +687,65 @@ pub trait UiClient: PollClient {
     }
 }
 
+pub struct ClientBuilder<B: 'static = Empty> {
```
nickray, line 690: `I` instead of `B`. It's a bit tricky to stay on top of Backend vs Id throught this code, I think this would help.

src/client.rs
```diff
@@ -685,6 +687,65 @@ pub trait UiClient: PollClient {
...
+    pub fn new(id: impl Into<PathBuf>) -> Self {
+        Self {
+            id: id.into(),
+            backends: &[],
+        }
+    }
+}
+
+impl<B: 'static> ClientBuilder<B> {
+    pub fn backends<C: 'static>(self, backends: &'static [BackendId<C>]) -> ClientBuilder<C> {
```
nickray, line 705: Perhaps `J` instead of `C`?

src/client.rs
```diff
@@ -685,6 +687,65 @@ pub trait UiClient: PollClient {
...
+impl ClientBuilder {
+    pub fn new(id: impl Into<PathBuf>) -> Self {
+        Self {
+            id: id.into(),
+            backends: &[],
+        }
+    }
+}
+
+impl<B: 'static> ClientBuilder<B> {
```
nickray, line 704: `I` -> `B`

src/pipe.rs
```diff
@@ -71,11 +72,12 @@ cfg_if::cfg_if! {
 // https://xenomai.org/documentation/xenomai-2.4/html/api/group__native__queue.html
 // https://doc.micrium.com/display/osiiidoc/Using+Message+Queues
 
-pub struct ServiceEndpoint {
+pub struct ServiceEndpoint<B: 'static = ()> {
```
nickray: `I`

nickray (changes requested):

robin-nitrokey:
Updated:
- Changed `B` to `I` for the backend ID (and `C` to `J`).
- Added `NoData` struct (though I’m not really convinced ;)).
- Added documentation.
- Changed “software” to “core” when referring to the default request implementation in Trussed.

robin-nitrokey:
Rebased onto 5c9bc60eaecf7d38b3512b973ac74309980ae721 and added to changelog.

src/types.rs
```diff
@@ -233,17 +241,36 @@ pub mod consent {
...
+        }
+    }
+}
+
+// The "CoreContext" struct is the closest equivalent to a PCB that Trussed
 // currently has. Trussed currently uses it to choose the client-specific
 // subtree in the filesystem (see docs in src/store.rs) and to maintain
 // the walker state of the directory traversal syscalls.
-pub struct ClientContext {
+pub struct CoreContext {
```
nickray, line 267: Shouldn't this be non-exhaustive in light of other discussions?

daringer:
:+1:

src/types.rs
robin-nitrokey: I’ll add it to #71.  It has to be part of the public API so that the backends have access to it.

PR #70, opened 2023-01-22
 Add API extensions 

robin-nitrokey:
This patch adds API extensions to Trussed.  To avoid having to add a
type parameter to Request and Reply, we serialize the requests and
replies.  Extensions can be implemented by custom backends and are
dispatched by the Backends implementation provided to the Service.  An
additional type parameter is needed for ClientImplementation to express
the supported extensions.  See tests/ext.rs for an example
implementation.

This implementation has two limitations:
- As we cannot define a non-generic object-safe trait to execute
  extensions, Backends cannot just return a reference to the extension
  <s>(as we do for the backends)</s>.  Instead it has to call the extension
  implementation directly.
  Edit: After adding backend context, we have to do the same for backends too.
- The serialization and deserialization of the requests and replies of
  course adds additional overhead.  An alternative would be to make
  Request and Reply generic over the set of extensions.

Based on: https://github.com/trussed-dev/trussed/pull/68


robin-nitrokey:
Still missing from this PR is a way to an extension to store data in the client context.

robin-nitrokey:
Added backend-specific context that can also be used by extensions.  The Backends implementation and dispatch is pretty verbose, but I think the overall approach works well.  If desired, macros could be used to reduce the repetition.

robin-nitrokey:
Rebased onto 6b6ee2132cfe1947b441dc9214de7ed551efa450.

szszszsz (changes requested):

src/api.rs
```diff
@@ -482,5 +488,9 @@ pub mod reply {
 
         WriteCertificate:
           - id: CertId
+
+        // reply size is chosen to be smaller than the largest standard syscall, Encrypt
+        Extension:
+          - reply: Bytes<{ MAX_MESSAGE_LENGTH + 2 * MAX_SHORT_DATA_LENGTH }>
```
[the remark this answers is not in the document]
szszszsz: Yes, that's the part I do not get. Why do you not want to serialize Encrypt, if you want to encrypt data using another backend?

src/api.rs
```diff
@@ -332,6 +334,10 @@ pub mod request {
           - location: Location
           - der: Message
 
+        // request size is chosen to be smaller than the largest standard syscall, Decrypt
+        Extension:
+          - id: u8
+          - request: Bytes<{ 2 * MAX_MESSAGE_LENGTH + 2 * MAX_SHORT_DATA_LENGTH }>
```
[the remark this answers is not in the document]
robin-nitrokey: The idea is not to be able to serialize the `Decrypt` request into this buffer.  We have to choose some arbitrary limit for the available buffer, and the idea is to use all the memory that we already need for one of the other variants, but not more to avoid adding overhead.

src/api.rs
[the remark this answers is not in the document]
robin-nitrokey: If a backend wants to implement the `Encrypt` syscall, it can just use the existing request implementation.  This variant is only relevant for API extensions, i. e. new syscalls that are only available in some backend.  For these syscalls, we use serialization to avoid adding a type parameter.  And to store the serialized data, we need a fixed buffer size.  It is chosen so that the enums do not grow larger than already needed, but the size could also be set to a different value.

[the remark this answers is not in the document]
robin-nitrokey: Ideally, the buffer size would be configured by the runner, but we can’t do that atm.

src/api.rs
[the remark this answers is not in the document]
robin-nitrokey: Updated the comment.

src/api.rs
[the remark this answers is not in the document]
robin-nitrokey: Updated the comment.

szszszsz (approved):
Good work!
Please add doc comments regarding:
- the potential failing of the Encrypt/Decrypt calls in the ExtensionAPI due to the serialization;
- why the size of the new type is chosen this way.

robin-nitrokey:
Rebased onto d32a309a9c73ec1db1813f190155eaf3a377f4a8.

src/ext.rs
```diff
@@ -0,0 +1,121 @@
...
+    platform::{Platform, Syscall},
+    postcard_deserialize, postcard_serialize_bytes,
+    service::ServiceResources,
+    types::ClientContext,
+};
+
+use serde::{de::DeserializeOwned, Serialize};
+
+pub trait Extension {
+    type Request: DeserializeOwned + Serialize;
```
sosthene-nitrokey, line 17: I think we want to have `type Request<'a>: Deserialize<'a> + Serialize`. Same for the Response type.

Otherwise we risk having high stack usage: Say a command comes from the host. The app would have to parse it's relevant data into a `Request` that looks like `struct Request { field1: Bytes<1000>, field2…}` which takes 1000 bytes on the stack. The same thing then happens in the backend. The same thing also happens with the reply.

By using the lifetimes I suggested we can instead use `struct Request<'a> { field1: &'a [u8], field2: …}`, which requires very little stack space.

src/ext.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    const ID: Self::Id;
+}
+
+pub trait ExtensionClient<E: Extension>: PollClient {
+    fn id() -> u8;
+
+    fn extension<Rq, Rp>(&mut self, request: Rq) -> ExtensionResult<'_, E, Rp, Self>
+    where
+        Rq: Into<E::Request>,
```
sosthene-nitrokey, line 57: `AsRef`?

src/ext.rs
sosthene-nitrokey: I realize this doesn't work with the current `Poll` trait that returns an owned reply. It does seem inefficient to always move the request/replies, given that they're pretty large.

I think returning references and let the client `clone` when necessary would be more efficient.

nickray:
This feels a bit unsatisfying to me. The real problem seems to me we want Request to be a trait, so it can be extended outside Trussed. But we also want an enum, so we can match on the Request.

I don't want to make purity the enemy of practicality, so not totally against merging this kind of "out of type system" approach, but it feels very much like C APIs of old and I think long term we can do better.

Does this sound right?

A related problem I have is with the registries we're building up (e.g. the assignment `WriteFile: 19`). We only need these for technical reasons, we'd really like to have the type encode the method instead of a random assignment that needs to be managed.

My ideal would be a way for Trussed upstream to maintain a bunch of core Requests, a bunch of extension requests, and clients to permissionlessly have a bunch of private Requests, that all get merged into an actual API at build time - without registries.

sosthene-nitrokey:
There's one advantage to actually using serde to serialize/deserialize requests/response. If we go the route of having the backend be isolated from the client with something like TrustZone, we actually can't use the current `Request`/`Reply` enums since the client apps would be assumed malicious and would be capable of placing invalid data in the shared memory region. In this case postcard serialization/deserialization makes sense.

sosthene-nitrokey:
If we want to replace the enum I'm thinking we could just have as a request/reply types a form of  `Box<dyn Any>` (though, heapless doesn't appear to provide a suitable replacement) and then have each "extension plugin" register the [`TypeId`](https://doc.rust-lang.org/core/any/struct.TypeId.html)s that are relevant to it. The dispatch would be based on the `TypeId`.

robin-nitrokey:
An alternative to implement extensions using the type system would be making the `Request` and `Reply` types generic over the extension set (or replacing them with a type parameter altogether) and having the runner define the available requests and dispatching them to Trussed or a backend.  (Ideally, this could be done by a macro so the user only has to define the extensions they want to support.)  This was the approach I initially wanted to explore, but given your reluctance to add new type parameters, I figured that an approach outside of the type system would be the only realistic way to move forward with this.

nickray:
> There's one advantage to actually using serde to serialize/deserialize requests/response. If we go the route of having the backend be isolated from the client with something like TrustZone, we actually can't use the current Request/Reply enums since the client apps would be assumed malicious and would be capable of placing invalid data in the shared memory region. In this case postcard serialization/deserialization makes sense.

To recall (you know this) the idea for TrustZone is that each client has its request/reply interchange in its own (untrusted) memory region. Yes, the client could maliciously put invalid data there, which deserialize would catch, but an overarching idea is to have "compile-time security" for Trussed apps, i.e., we would do code analysis to ensure the apps don't request with invalid data.

Downside of serialization is a) more memory use and b) more processing.

I understand that these arguments might reach their limits, for now I'm still pretty fond of zerocopy/zeroserde.

In the long term, maybe the right answer is "why not both", i.e., formalize "Service <-> Client transport" and offer multiple implementations with their respective tradeoffs.

> If we want to replace the enum I'm thinking we could just have as a request/reply types a form of Box<dyn Any> (though, heapless doesn't appear to provide a suitable replacement) and then have each "extension plugin" register the [TypeId](https://doc.rust-lang.org/core/any/struct.TypeId.html)s that are relevant to it. The dispatch would be based on the TypeId.

That's essentially what dtolnay's approach does if I remember correctly. We would probably need our own Box-ish type (or some kind of `&'static` like `linkme`) to respect what we need for the hypothetical TrustZone extension.

> An alternative approach to implement extensions using the type system would be making the Request and Reply types generic over the extension set (or replacing them with a type parameter altogether) and having the runner define the available requests and dispatching them to Trussed or a backend. (Ideally, this could be done by a macro so the user only has to define the extensions they want to support.) This was the approach I initially wanted to explore, but given your reluctance to add new type parameters, I figured that an approach outside of the type system would be the only realistic way to move forward with this.

Making the Trussed API "flexible + permissionless extendable, type-safe with compile-time security" would be a good argument for new generic types :)

`linkme`'s reliance on the linker is maybe too much magic, but also maybe not if it gets rid of "macros by example".

Generally we haven't explored proc-macros at all, either.

robin-nitrokey:
Updated:
- Rebased onto c014d0a29e4a8e464febdfd13c009804e21ab8de.
- Moved behind `ext` feature flag.  Extensions are extracted from the `Dispatch` trait and require a new `ExtensionDispatch` trait.  The main implication of this is that we need a helper trait, `DispatchRequirement`, to determine the traits required by `Service`.  Otherwise the changes caused by the feature are only local.

Cargo.toml
```diff
@@ -57,6 +57,7 @@ once_cell = "1.13.0"
 
 [features]
 default = ["default-mechanisms", "clients-5"]
+ext = []
```
nickray, line 60: This is too short and C (same actually for `virt`), perhaps `api-extensions`?

Cargo.toml
```diff
@@ -120,5 +121,5 @@ test-attestation-cert-ids = []
 # interchange = { git = "https://github.com/trussed-dev/interchange", branch = "main" }
 
 [package.metadata.docs.rs]
-features = ["virt"]
+features = ["ext", "virt"]
```
nickray, line 124: I'm not even 100% what "virt" is. A `mock-platform`? As it's not just running in RAM, but also using the hardcoded seed of 42s. Probably we should decouple this anyway, so we have a builder with a real `CryptoCoreRng`, one passing a seed, and perhaps for ultimate unsafe convenience one with a hardcoded seed.

src/ext.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use serde::{de::DeserializeOwned, Serialize};
+
+pub trait Extension {
+    type Request: DeserializeOwned + Serialize;
+    type Reply: DeserializeOwned + Serialize;
+}
+
+/// Dispatches extension requests to custom backends.
+pub trait ExtensionDispatch<P: Platform>: Dispatch<P> {
```
nickray, line 22: (The following is based on the feeling that cfg-if'ing `DispatchRequirement` leads to too much of a "parallel worlds" split of upstream Trussed)

I was thinking this would look more like:

```rust
pub trait Dispatch<P> {
    /// The ID type for the custom backends used by this dispatch implementation.
    type BackendId: 'static;
    /// The context type used by this dispatch.
    type Context: Default;
    /// The ID type for the extensions supported by this dispatch implementation.
    #[cfg(feature = "api-extensions")]
    type ExtensionId: TryFrom<u8, Error = Error>;

    fn request(
        &mut self,
        backend: &Self::BackendId,
        ctx: &mut Context<Self::Context>,
        request: &Request,
        resources: &mut ServiceResources<P>,
    ) -> Result<Reply, Error> { ... }

    #[cfg(feature = "api-extensions")]
    fn extension_request(
        &mut self,
        backend: &Self::BackendId,
        extension: &Self::ExtensionId,
        ctx: &mut Context<Self::Context>,
        request: &request::Extension,
        resources: &mut ServiceResources<P>,
    ) -> Result<reply::Extension, Error> { ... }
}
```

Might not be so nice because 2532-associated-type-defaults is unstable (for ExtensionId).
On the other hand, Trussed doesn't really semantically understand ExtensionId anyway, so it could just pass it along untouched:

```rust
pub trait Dispatch<P> {
    /// The ID type for the custom backends used by this dispatch implementation.
    type BackendId: 'static;
    /// The context type used by this dispatch.
    type Context: Default;

    fn request(
        &mut self,
        backend: &Self::BackendId,
        ctx: &mut Context<Self::Context>,
        request: &Request,
        resources: &mut ServiceResources<P>,
    ) -> Result<Reply, Error> { ... }

    #[cfg(feature = "api-extensions")]
    fn extension_request(
        &mut self,
        backend: &Self::BackendId,
        ctx: &mut Context<Self::Context>,
        // extension: u8,
        request: &request::Extension,
        resources: &mut ServiceResources<P>,
    ) -> Result<reply::Extension, Error> { 
        let _ = (backend, /*extension,*/ ctx, request, resources);
        Err(Error::RequestNotAvailable)
    }
}
```

I commented out the `extension: u8` as the `request::Extension` already contains it. But now at this point, there isn't an actual difference between `fn request` and `fn extension_request`, so... The entire `api-extensions` feature is seems moot and we can just stick with a single `request` method?

I think you're trying to add some type safety via the `ExtensionId`, but since in `request::Extension` you're actually passing the naked `u8` with no indication of its type, it seems this gives no additional safety anyway?

In which case (while staying with this simple approach, and assuming you have desire for multiple different ExtensionId types), perhaps that API would be better as 
```
        Extension:
          - id: Bytes<{ some small number }>,  // <- serializing more metadata
          - request: Bytes<{ crate::config::EXTENSION_REQUEST_LENGTH }>
```
or even
```
Extension:
  kind: u8,  // <- indicating the underlying type implementing ExtensionId
  id: u8,
  request: Bytes
```

src/service.rs
```diff
@@ -570,6 +624,11 @@ impl<P: Platform> ServiceResources<P> {
                     .map(|id| Reply::WriteCertificate(reply::WriteCertificate { id } ))
             }
 
+            #[cfg(feature = "ext")]
```
nickray, line 627: This would just be non-feature flagged.

src/ext.rs
nickray: The above is for the backend + service side of things, I do think it would be nice to have upstream code factoring out ergonomic and type safe-er client struct APIs.

src/service.rs
```diff
@@ -67,57 +85,93 @@ impl<P: Platform> ServiceResources<P> {
...
+        &mut self,
+        dispatch: &mut D,
+        backend: &BackendId<D::BackendId>,
+        ctx: &mut Context<D::Context>,
+        request: &Request,
+    ) -> Result<Reply, Error> {
+        #[cfg(feature = "ext")]
+        if let Request::Extension(request) = &request {
+            if let BackendId::Custom(backend) = backend {
+                return D::ExtensionId::try_from(request.id)
```
nickray, line 148: Unless I'm missing something, I think this entire PR is "just" factoring this `D::ExtensionId::try_from(request.id)` call from the actual backend upstream to the service.

src/ext.rs
```diff
@@ -0,0 +1,145 @@
...
+pub trait ExtensionId<E> {
+    type Id: Into<u8>;
+
+    const ID: Self::Id;
+}
+
+pub trait ExtensionClient<E: Extension>: PollClient {
+    fn id() -> u8;
+
+    fn extension<Rq, Rp>(&mut self, request: Rq) -> ExtensionResult<'_, E, Rp, Self>
```
nickray, line 79: I think this could be done more elegantly, with less code, by using the `pub trait Extension` above (with a little polish like @sosthene-nitrokey's comments on Deserialize<'a>) and adding the method directly to `PollClient`, roughly as follows:

```rust
pub type ClientResult<'c, T, C> = core::result::Result<FutureResult<'c, T, C>, ClientError>;
pub type ClientExtensionResult<'c, T, C> = core::result::Result<FutureExtensionResult<'c, T, C>, ClientError>;

pub trait PollClient {
    fn request<T: From<crate::api::Reply>>(&mut self, req: impl Into<Request>) -> ClientResult<'_, T, Self>;
    fn poll(&mut self) -> core::task::Poll<core::result::Result<Reply, Error>>;
    fn syscall(&mut self);
    // new
    fn extension<E: Extension>(&mut self, req: E::Request) -> ClientExtensionResult<'_, E::Reply, Self>;
}
````
I know this is starting to contradict the ask to feature-flag extensions, but given we are including the "raw" Extension request in api.rs, if we can support handling it with minimal code we might as well keep it.
Then all the `PollClient::extension` has to do is serialize `E::Request` into the extension variant, and continue like `PollClient::poll`.

I even think we might be able to merge the two future result and client result types with suitable conversion traits. Or at the very least, we'd have not too much code to feature flag.

**Review by nickray — CHANGES_REQUESTED** [review]

**Comment by nickray** [comment]
I'm sorry! I made a bit of a mess with this review.

Concretely I suggest these changes: https://github.com/robin-nitrokey/trussed/compare/api-ext...nickray:trussed:api-ext

This means:
- like for RSA, we register the `SerdeExtensions: 0x5E` code point upstream (code 0x5E chosen as 0xff tends to be super-special, and it looks like SE)
- no `DispatchRequirement`, just `Dispatch`
-  `try_from(extension_id)` is removed from backend/service, but not fully punted to custom backends, instead `src/serialized_extensions.rs` implements it, keeping the `Impl` trait usefully ergonomic.

**Comment by robin-nitrokey** [comment]
I’m not a big fan of dropping the extension type parameter from `ExtensionClient`.  This was a way to ensure that all extensions required by an application are provided by the runner.  For example, the auth extension would use:

```
trait AuthClient: ExtensionClient<AuthExtension> { ... }
```

Applications could then use this trait as a bound.  Failing to provide an extension in the runner would cause a compiler error.  

Dropping this parameter means that the application will compile even if the extension is not provided, making it harder to maintain applications and runners.  This is especially bad if an application has a feature flag to enable support for an extension.  It would be hard to ensure in CI that the default feature set does not use the extension.

File: Cargo.toml (hunk 5 above)
Comment (robin-nitrokey) [reply Cargo.toml]: NB: We chose `virt` because `virtual` is a keyword and `virtual_platform` seemed excessive, but if you have a better idea, I’ll change it.

File: Cargo.toml (hunk 6 above)
Comment (robin-nitrokey) [reply Cargo.toml]: The hardcoded seed dates back to this commit:
https://github.com/trussed-dev/trussed/commit/87e0c64ad94af978012cf9401bbeb5d65c29528e#diff-2df46fbc46e4818d054b9537e84ce0a18d7c95064657a2422040ac9d30e96914

I did not have the time and need to investigate this yet, but making this configurable would of course be better.

**Comment by nickray** [comment]
> I’m not a big fan of dropping the extension type parameter from `ExtensionClient`. This was a way to ensure that all extensions required by an application are provided by the runner.

How would you do this if you needed two or more extensions?

In any case, if we bring back the type parameter, then I think we need it in `ClientImplementation` again - or `ExtensionClient` actually wraps it instead of extending its methods.

Generally I thought it was cleaner not to have an unused generic parameter in the "upstream" client implementation; if we prefer implementing traits to wrapping it though, we could designate it as a "generic" extension point for use by out-of-repo libraries to use. Although then what happens if two want to use it? So you see, all in all I feel adding it for use in serde-extensions (with a single "set" of Extensions) s a bit of a tight coupling.

**Comment by robin-nitrokey** [comment]
> How would you do this if you needed two or more extensions?

The traits would be additive.  Let’s say we have two extensions, `A` and `B`.  The client would implement `ExtensionClient<A>` and `ExtensionClient<B>`.  The extensions would provide the `AClient: ExtensionClient<A>` and `BClient: ExtensionClient<B>` traits.  Applications could be generic over `C: AClient + BClient`, similar to the existing client traits.

> Generally I thought it was cleaner not to have an unused generic parameter in the "upstream" client implementation; if we prefer implementing traits to wrapping it though, we could designate it as a "generic" extension point for use by out-of-repo libraries to use. Although then what happens if two want to use it? So you see, all in all I feel adding it for use in serde-extensions (with a single "set" of Extensions) s a bit of a tight coupling.

The type parameter is the `Dispatch` implementation.  Currently, we only use it to check the serde extensions provided by the dispatch using the `ExtensionId` trait, but that concept could be extended for any other extension mechanism (or even any other feature that depends on the dispatch provided by the runner).  So I see this as a very general solution.

> My vote I think would go for wrapping (impl PollClient, Extension) -> impl PollClient, what do you think?

Can you elaborate?  I’m not sure I understand what you mean by that.

File: Cargo.toml [hunk 11]
```diff
@@ -121,5 +121,5 @@ test-attestation-cert-ids = []
 # interchange = { git = "https://github.com/trussed-dev/interchange", branch = "main" }
 
 [package.metadata.docs.rs]
-features = ["ext", "virt"]
+features = ["serialized-extensions", "virt"]
```
robin-nitrokey, line 124: typo: `serde-extensions`

nickray: Thanks, I have that on my upstream branch (and didn't intend to push that on top of your changes to this without prior discussion).

nickray:
Reverted my last commit and extended the test instead to convince myself of how things work :)

Apart from implementing `ExtensionDispatch` being pretty repetitive (which can be improved later), I'm pretty happy with this PR now, so if you agree, let's merge?

robin-nitrokey:
Thanks!  I’ve added some basic doc comments for `serde_extensions`.  I think we can merge now.

nickray:
Woohoo, habemus extensions.

nickray (approved):

Issue #62, opened 2023-01-17
Reduce potential for breaking changes in public API


robin-nitrokey:
With the next breaking release, we should take some precautions to reduce potential breaking changes in the future.  For example:
- make enums like `Error`, `Request`, `Reply` and `Mechanism` non-exhaustive
- remove unnecessary re-exports like cbor and postcard serialization

Issue #63, opened 2023-01-17
Fork or replace postcard


robin-nitrokey:
Upgrading postcard from 0.7 to 1.0 would be a breaking change and invalidate existing FIDO credentials generated with `fido-authenticator`.  To avoid outdated dependencies, we should consider forking or replacing postcard.

nickray:
If we replace, a possible alternative would be bincode v2 (unreleased), given its widespread use outside embedded.

daringer:
This is less a technical question than a migration/user question first. I don't think it is an option to invalidate all FIDO credentials with a firmware update. So we first need a migration strategy, right? My take on this would be:

* existing credentials should still use postcard 0.7 
* new credentials should use <insert-new-fancy-serializer> 
* during firmware update (or some other comm way), tell the user that she/should re-register u2f credentials 
* remove old postcard after a reasonable migration time

So far I understand, forking postcard would be the only option if we would not like to do this, right? 

From the technical point of view, I don't really have an opinion, but [benchmarks](https://github.com/only-cliches/noproto#benchmarks) seem to swing in favor of `bincode`.

robin-nitrokey:
As far as I see, postcard deserialization is only used once:  to deserialize `reply::Encrypt`.  My first approach would be see if we can just manually deserialize this struct.

https://github.com/trussed-dev/trussed/blob/8e347abf99dba81fca11c5779a473d263a7b0565/src/mechanisms/chacha8poly1305.rs#L210-L214
https://github.com/trussed-dev/trussed/blob/8e347abf99dba81fca11c5779a473d263a7b0565/src/api.rs#L382-L385

robin-nitrokey:
Here is a quick draft of how a manual implementation could look like: https://github.com/robin-nitrokey/trussed/commit/11429b7789573f590abf96c0ce6cc28a7eec047b  It’s not trivial, but not too bad either – 50 lines of code that are rather easy to review and to test, no external dependencies.

nickray:
Nice - this amount of code seems reasonable to keep around for a rather long "reasonable migration time".

PR #71, opened 2023-01-23
Make public API more stable

robin-nitrokey:
This PR contains two changes that make the public API more robust:
- Marking enums like `Error`, `Response`, `Reply` and `Mechanism` as non-exhaustive makes it possible to add features without breaking compatibility.
- Removing re-exports of the serialization functions from `cbor_smol` and `postcard` gives us more flexibility to change these implementation details.

Fixes https://github.com/trussed-dev/trussed/issues/62

### DIFF
--- a/src/api/macros.rs
+++ b/src/api/macros.rs
@@ -3,6 +3,7 @@ macro_rules! generate_enums {
 
     #[derive(Clone, Eq, PartialEq, Debug)]
     #[allow(clippy::large_enum_variant)]
+    #[non_exhaustive]
     pub enum Request {
         DummyRequest, // for testing
         $(
@@ -13,6 +14,7 @@ macro_rules! generate_enums {
 
     #[derive(Clone, Eq, PartialEq, Debug)]
     #[allow(clippy::large_enum_variant)]
+    #[non_exhaustive]
     pub enum Reply {
         DummyReply, // for testing
         $(
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,11 +48,12 @@ pub use error::Error;
 pub use platform::Platform;
 pub use service::Service;
 
-pub use cbor_smol::{cbor_deserialize, cbor_serialize, cbor_serialize_bytes};
+pub use cbor_smol::{cbor_deserialize, cbor_serialize_bytes};
 pub use heapless_bytes::Bytes;
-pub use postcard::{from_bytes as postcard_deserialize, to_slice as postcard_serialize};
 
-pub fn postcard_serialize_bytes<T: serde::Serialize, const N: usize>(
+pub(crate) use postcard::from_bytes as postcard_deserialize;
+
+pub(crate) fn postcard_serialize_bytes<T: serde::Serialize, const N: usize>(
     object: &T,
 ) -> postcard::Result<Bytes<N>> {
     let vec = postcard::to_vec(object)?;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated `littlefs2` to 0.4.0.
 - Made `Request`, `Reply`, `Error`, `Context`, `CoreContext`, `Mechanism`,
   `ui::Status` non-exhaustive.
+- Made `postcard_deserialize`, `postcard_serialize` and
+  `postcard_serialize_bytes` private.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Upgrade the `interchange` dependency to version 0.3.0 ([#99][])
   - As a consequence the type `pipe::TrussedInterchange` becomes a const`pipe::TRUSSED_INTERCHANGE`
 - Updated `littlefs2` to 0.4.0.
+- Made `Request`, `Reply`, `Error`, `Context`, `CoreContext`, `Mechanism`,
+  `ui::Status` non-exhaustive.
 
 ### Fixed
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,6 +7,7 @@ pub type Result<T, E = Error> = core::result::Result<T, E>;
 
 #[derive(Copy, Clone, Eq, PartialEq, Debug)]
 #[repr(u32)]
+#[non_exhaustive]
 pub enum Error {
     // cryptoki errors
     HostMemory = 0x0000_0002,
--- a/src/types.rs
+++ b/src/types.rs
@@ -170,6 +170,7 @@ pub mod ui {
     // TODO: Consider whether a simple "language" to specify "patterns"
     // makes sense, vs. "semantic" indications with platform-specific implementation
     #[derive(Copy, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+    #[non_exhaustive]
     pub enum Status {
         Idle,
         WaitingForUserPresence,
@@ -242,6 +243,7 @@ pub mod consent {
 /// The context stores the state used by the standard syscall implementations, see
 /// [`CoreContext`][].  Additionally, backends can define a custom context for their syscall
 /// implementations.
+#[non_exhaustive]
 pub struct Context<B> {
     pub core: CoreContext,
     pub backends: B,
@@ -260,6 +262,7 @@ impl<B: Default> From<CoreContext> for Context<B> {
 // currently has. Trussed currently uses it to choose the client-specific
 // subtree in the filesystem (see docs in src/store.rs) and to maintain
 // the walker state of the directory traversal syscalls.
+#[non_exhaustive]
 pub struct CoreContext {
     pub path: PathBuf,
     pub read_dir_state: Option<ReadDirState>,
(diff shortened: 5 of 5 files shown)


robin-nitrokey:
Previously we talked about making [`types::ui::Status`](https://docs.rs/trussed/latest/trussed/types/ui/enum.Status.html) non-exhaustive.  I don’t think this would be appropriate – how should runners handle unknown states?

robin-nitrokey:
Added the CBOR re-exports again because they are used in the `serialize_key` syscalls.  Still not sure if we should really have them, but dropping postcard is more important.

robin-nitrokey:
> I would vote for [types::ui::Status](https://docs.rs/trussed/latest/trussed/types/ui/enum.Status.html) handling inside the runners should have a fallback `_ => foo()` otherwise runners would have to be adapted for any [types::ui::Status](https://docs.rs/trussed/latest/trussed/types/ui/enum.Status.html) updates, which as of today is LED only, but sooner or later might also include other interactions with the user/hardware...

But can we really model these more complex states inside Trussed?  Or would we need a way to give the runner more control over the UI state machine?

daringer:
Good point, but in consequence this would mean we've to come up with a extension mechanism? I would prefer a simple solution, which would work for simple cases now (fallback?) and once the need arises we would look at a more flexible solution. Anyways not a high prio topic, so I would add it to the discussion list for the next steering meeting...

src/lib.rs
```diff
@@ -42,11 +42,12 @@ pub use error::Error;
...
 
-pub use cbor_smol::{cbor_deserialize, cbor_serialize, cbor_serialize_bytes};
+pub use cbor_smol::{cbor_deserialize, cbor_serialize_bytes};
 pub use heapless_bytes::Bytes;
-pub use postcard::{from_bytes as postcard_deserialize, to_slice as postcard_serialize};
 
-pub fn postcard_serialize_bytes<T: serde::Serialize, const N: usize>(
+pub(crate) use postcard::from_bytes as postcard_deserialize;
+
+pub(crate) fn postcard_serialize_bytes<T: serde::Serialize, const N: usize>(
```
sosthene-nitrokey, line 50: In the RSA PR, I used postcard in the import API to serialize structured data as bytes for key import. Given that it makes sense for me to keep postcard as part of the public API.
Maybe it could be kept private if we make RSA key import it's own method were the Client makes the serialization transparently.

src/api/macros.rs
```diff
@@ -12,6 +13,7 @@ macro_rules! generate_enums {
 
     #[derive(Clone, Eq, PartialEq, Debug)]
     #[allow(clippy::large_enum_variant)]
+    #[non_exhaustive]
```
sosthene-nitrokey, line 16: I think the `impl_request` and  `impl_reply` should also be marked `non_exhaustive` right?

sosthene-nitrokey (commented):
It might be a much bigger change but I think we should also consider switching to a builder pattern for making requests. It could easily be generated by the macros. This would allow adding fields to any Request without any breakage.

src/lib.rs
robin-nitrokey: Hmm, I’ll have to take another look at that use case.  My main motivation for dropping the `postcard` re-export is making https://github.com/trussed-dev/trussed/issues/63 feasible.

robin-nitrokey:
> It might be a much bigger change but I think we should also consider switching to a builder pattern for making requests. It could easily be generated by the macros. This would allow adding fields to any Request without any breakage.

Hmm, and how would the `syscall` interface look like?  My best guess would be something like this:

- no arguments
  ```
  syscall!(Uptime);
  ```
- only required arguments:
  ```
  syscall!(ReadFile, (Location::Internal, path));
  ```
- required and optional arguments:
  ```
  syscall!(WriteFile, (Location::Internal, path, data), |r| r.user_attribute(UserAttribute));
  ```
We could also discuss grouping the requests by topic as we do for the `Client` traits, and enforcing the mapping between requests and replies in the type system.

Anyway, I think we should discuss this separately as it is a much more invasive and fundamental change.  What do you think about moving this topic into a discussion?

src/api/macros.rs
robin-nitrokey: Hmm, I’m not sure.  Without the fundamental changes to the requests that you suggested, I think adding fields to `Request` or `Response` should be breaking.

sosthene-nitrokey:
No, I meant generating the builder types from the `impl_reply` and `impl_request` macros. Then we would keep the current `syscall` macro and have something like:

```rust
syscall!(client.request(Request::read_file(mandatory, arguments).optional_arg(smth).build()));
```

Other variants:

```rust
syscall!(client.read_file(mandatory, arguments).optional_arg(smth))

syscall!(Request::read_file(mandatory, arguments).optional_arg(smth).build(client))
```

sosthene-nitrokey:
Or getting rid of the `syscall` macro, even currently I don't see why it need to be a macro and not a function:

```rust
Request::read_file(mandatory, arguments).optional_arg(smth).syscall(client);
Request::read_file(mandatory, arguments).optional_arg(smth).try_syscall(client);
```

This could exist in parallel to the current way to build syscalls to migrate progressively.

daringer:
I am a big fan of this:
```
Request::read_file(mandatory, arguments).optional_arg(smth).syscall(client);
```
but I would also be in favor of having this separated from this PR as:
* this could be purely optional (as @sosthene-nitrokey suggests)
* this could also be a trigger to talk about `client.rs` which is also highly repetitive & redundant

nickray:
You guys are on a roll, I love it <3

For the file operations, my ideal would be for the clients to have something closely resembling `std::fs`/`std::path` as API, appropriately channeled over the generic Response/Request mechanism.

The `syscall!` was an early hack that never got revisited.
At first glance, I feel like the `client.<method>(args){,.optional(args)}.call()` seems a bit more idiomatic and discoverable than the `Request:<method>(args){,optional(args}.call(client)` approach (although the former is probably syntactic sugar for the latter  in reality). As then `Client` can be bound on traits requiring the subset of methods a Trussed app requires, and documentation more clearly splits up the methods.

Regarding UI, general thoughts are that the apps should rely on semantic signals (such as the user is present, or the user is identified... and probably extendable once you have touch screens or whatnot), whereas the runner should be able to fully inject the UI implementation. So might make sense to make a list of envisioned UI elements (LED, button, screen, pin pad) and figure out the semantic methods and states from an app perspective, and how to wire this up into Trussed without Trussed service being aware of the implementations.

sosthene-nitrokey:
> The `syscall!` was an early hack that never got revisited. At first glance, I feel like the `client.<method>(args){,.optional(args)}.call()` seems a bit more idiomatic and discoverable than the `Request::<method>(args){,optional(args}.call(client)` approach (although the former is probably syntactic sugar for the latter in reality). As then `Client` can be bound on traits requiring the subset of methods a Trussed app requires, and documentation more clearly splits up the methods. The main goal is that writing app code feels "normal" to a non-embedded developer, even maybe a bit like Python. They want to (I think) do "simple things" like `let key = client.generate_p256_key();`, but have the power to do more complex things if needed.
> 
> Challenge: Could we in some way replace `let x = client.method().optional().do()` get rid of the final do? As in some clever trick involving ephemeral objects? It would be annoying for the above keygen to turn into `let key = client.generate_key().<some_name>()`.

Currently I don't think so. If we end up going the `async` way once async traits land (It looks like `RTIC` is going this way) we use the `IntoFuture` trait to have something like:

```rust
let x = client.method().optional().await;
```

I think removing any `call` method would be confusing though.

.github/workflows/firmwares.yml
```diff
@@ -59,6 +59,7 @@ jobs:
         run: | 
           echo "[patch.crates-io]" >> solo2/runners/lpc55/Cargo.toml
           echo "trussed = { path = '../../../trussed' }" >> solo2/runners/lpc55/Cargo.toml 
+          echo "oath-authenticator = { git = 'https://github.com/robin-nitrokey/oath-authenticator', rev = '9340340814fa26c28e7ae352932f8f5462c8d7dc' }" >> solo2/runners/lpc55/Cargo.toml
```
nickray, line 62: To be fixed before merge.

robin-nitrokey:
Rebased onto 9fc19904b5970c421d81fbf28f63c13943a445b6 and made `Context` and `CoreContext` non-exhaustive.

daringer:
`ui:Status` as `non-exhaustive` is also to be added

.github/workflows/firmwares.yml
robin-nitrokey: Replaced with upstream.

robin-nitrokey:
Updated:
- Made `ui::Status` non-exhaustive.
- Added changelog entries.
- Replaced `oath-authenticator` patch with upstream.

daringer (approved):

.github/workflows/firmwares.yml
```diff
@@ -56,6 +56,7 @@ jobs:
         run: | 
           echo "[patch.crates-io]" >> solo2/runners/lpc55/Cargo.toml
           echo "trussed = { path = '../../../trussed' }" >> solo2/runners/lpc55/Cargo.toml 
+          echo "oath-authenticator = { git = 'https://github.com/trussed-dev/oath-authenticator', rev = '20d9018a567e436f9037d38985f6c0f96d9921fe' }" >> solo2/runners/lpc55/Cargo.toml
```
nickray, line 59: Fix before merge?

robin-nitrokey: It’s already upstreamed.  Should we just remove the patch here and allow failures on the firmwares build?

PR #75, opened 2023-01-25
Trigger syscall in PollClient::request

robin-nitrokey:
We can trigger the syscall as soon as we placed the request in the interchange.  Doing that directly in PollClient::request instead of calling PollClient::syscall afterwards makes the code a lot simpler.


sosthene-nitrokey (approved):
LGTM. Removing `PollClient::syscall` is a breaking change but we don't seem to be using it anywhere

PR #76, opened 2023-01-25
Make all features compile

robin-nitrokey:
This patch fixes compilation issues in non-default features so that we can compile Trussed with `--all-features`.


sosthene-nitrokey (changes requested):
It would maybe be good to add a `cargo check --all-targets --all-features` to the CI.

daringer (approved):

PR #78, opened 2023-01-30
Increase size of file writing

sosthene-nitrokey:
The increased size of files is a requirement for  PIV, 1kB certificates are too small. This commit also removes the `user_attribute` field that was never read anyway. This avoids increasing the total size of the interchange

Both changes are breaking changes:
- Changing the messages size for the read and write requests can cause compilation error for code that depends on the length of the `heapless` arrays
- Removing `user_attribute` will require changing every call to the `read_file` and `write_methods` to remove the `None`.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -279,8 +279,8 @@ pub mod request {
         WriteFile:
           - location: Location
           - path: PathBuf
-          - data: Message
           - user_attribute: Option<UserAttribute>
+          - data: LargeMessage
 
         UnsafeInjectKey:
           - mechanism: Mechanism        // -> implies key type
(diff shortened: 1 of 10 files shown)


nickray:
Yeah, we can remove `UserAttribute`. If I remember correctly, it would have allowed filtering a tree of files without loading the files themselves, which might be an optimization for apps with a large amount of such files. So since we don't use it, can add back later if and only if someone actually needs this.

sosthene-nitrokey:
This will we replaced by the streaming api. Current advances can be found in https://github.com/Nitrokey/trussed/pull/18. Once #96 and https://github.com/trussed-dev/littlefs2/pull/34 are merged I will upstream it.

Issue #84, opened 2023-01-31
Dangerous `unsafe` in `impl_reply`


sosthene-nitrokey:
https://github.com/trussed-dev/trussed/blob/main/src/api/macros.rs#L96

```rust
    impl From<Reply> for $reply {
        fn from(reply: Reply) -> reply::$reply {
            match reply {
                Reply::$reply(reply) => reply,
                _ => { unsafe { unreachable_unchecked() } }
            }
        }
    }
```

This `unsafe` is not sound and this impl is part of the public API. I think we could replace it with a `panic!`, since it is only used by [`PollClient::request`](https://docs.rs/trussed/latest/trussed/client/trait.PollClient.html#tymethod.request). To reduce the risk of panicking I would instead use `TryFrom`.

robin-nitrokey:
Yeah, this should really be a `TryFrom`.  If we make changes to the request and reply types, it could make sense to have a trait that defines the request and reply types for a syscall.

robin-nitrokey:
I’ll prepare a PR.  Will also be useful for API extensions too because I’m using `TryFrom` there, so I had to duplicate `FutureResult`.

sosthene-nitrokey:
I'm making a PR that adds a trait that does bindings between the request and reply types.

PR #85, opened 2023-01-31
Replace From<Reply> with TryFrom<Reply> in impl_reply

robin-nitrokey:
Previously, we would just trigger undefined behavior (core::hint::unreachable_unchecked) if there was a mismatch between the request and reply types for a Trussed syscall.  With this patch, we return an error instead.

Fixes https://github.com/trussed-dev/trussed/issues/84


sosthene-nitrokey:
I made PR that added trait to avoid confusion of replies and requests types with each other. See #86

robin-nitrokey:
See https://github.com/trussed-dev/trussed/pull/86

PR #86, opened 2023-01-31
Introduce type-based linking of requests and replies

sosthene-nitrokey:
This ensures that `PollClient::request` always matches the correct Request and Reply

This patch also fixes https://github.com/trussed-dev/trussed/issues/84 and replaces the `From` that used `unreachable_unchecked` with `TryFrom`.

The main additions are the traits:

```rust
pub trait RequestData: Into<Request> + TryFrom<Request, Error = crate::Error> {
    type CorrespondingReply: ReplyData<CorrespondingRequest = Self>;
}

pub trait ReplyData: Into<Reply> + TryFrom<Reply, Error = crate::Error> {
    type CorrespondingRequest: RequestData<CorrespondingReply = Self>;
}
```

And PollClient's `request` becomes 
```rust
// old
fn request<T: From<Reply>>(&mut self, req: impl Into<Request>) -> ClientResult<'_, T, Self>;
 // new
fn request<Rq: RequestData>(
      &mut self,
      req: Rq,
) -> ClientResult<'_, Rq::CorrespondingReply, Self>;
```

This also showed that the reply type for the `RemoveDir` was incorrect (it returned `RemoveDirAll`) which actually triggered UB is mentionned in #84

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -105,11 +105,11 @@ generate_enums! {
 }
 
 pub trait RequestData: Into<Request> + TryFrom<Request, Error = crate::Error> {
-    type CorrespondingReply: ReplyData<CorrespondingRequest = Self>;
+    type Reply: ReplyData<Request = Self>;
 }
 
 pub trait ReplyData: Into<Reply> + TryFrom<Reply, Error = crate::Error> {
-    type CorrespondingRequest: RequestData<CorrespondingReply = Self>;
+    type Request: RequestData<Reply = Self>;
 }
 
 pub mod request {
--- a/src/api.rs
+++ b/src/api.rs
@@ -104,12 +104,12 @@ generate_enums! {
     DebugDumpStore: 0x79
 }
 
-pub trait RequestData: Into<Request> + TryFrom<Request, Error = crate::Error> {
-    type Reply: ReplyData<Request = Self>;
+pub trait RequestVariant: Into<Request> + TryFrom<Request, Error = crate::Error> {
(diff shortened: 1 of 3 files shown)


sosthene-nitrokey:
I wonder if this checking could also be used on the backend side.

robin-nitrokey:
What do you think about dropping the `Corresponding` prefix?  Would be a bit easier to read.

sosthene-nitrokey:
It gets confusing to have multiple types named `Request` and `Reply`, but I agree it is a bit verbose. I initially added it because as I was dealing with the errors that came with the refactor it was confusing.

I removed it in79ccc04f8e07905ea5e9f9c02a04363158b114f7

robin-nitrokey:
I generally agree, but for associated types I think re-using existing names is fine (e. g. `TryFrom::Error`).

nickray:
This is super cool, it always bugged me that these types were not matched on a type level.

robin-nitrokey (approved):

src/api.rs
```diff
@@ -105,6 +104,14 @@ generate_enums! {
     DebugDumpStore: 0x79
 }
 
+pub trait RequestData: Into<Request> + TryFrom<Request, Error = crate::Error> {
```
nickray, line 107: `RequestVariant` and `ReplyVariant`? Would more precisely name what we expect.

sosthene-nitrokey: done in ef3deac6744a6ec231a7cbd366fbbce19f4d84d0

PR #89, opened 2023-02-03
Add RSA mechanisms and key serialization format

sosthene-nitrokey:
We will port the RSA implementation to a custom backend. While backends can have extensions, they can't have custom `Mechanisms` and `Serialization` for the normal APIs. This patch adds them so that they can be implemented by the custom backend.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -13,9 +13,6 @@ pub type MAX_OBJECT_HANDLES = consts::U16;
 pub type MAX_LABEL_LENGTH = consts::U256;
 pub const MAX_MEDIUM_DATA_LENGTH: usize = 256;
 pub type MAX_PATH_LENGTH = consts::U256;
-pub const MAX_KEY_MATERIAL_LENGTH: usize = 128;
-// must be above + 4
-pub const MAX_SERIALIZED_KEY_LENGTH: usize = 132;
 cfg_if::cfg_if! {
     if #[cfg(feature = "clients-12")] {
         pub type MAX_SERVICE_CLIENTS = consts::U12;
@@ -44,7 +41,14 @@ cfg_if::cfg_if! {
     }
 }
 pub const MAX_SHORT_DATA_LENGTH: usize = 128;
-pub const MAX_SIGNATURE_LENGTH: usize = 72;
+
+pub const MAX_SIGNATURE_LENGTH: usize = 512 * 2;
+// FIXME: Value from https://stackoverflow.com/questions/5403808/private-key-length-bytes for Rsa2048 Private key
+pub const MAX_KEY_MATERIAL_LENGTH: usize = 1160 * 2 + 72;
+
+// must be MAX_KEY_MATERIAL_LENGTH + 4
+pub const MAX_SERIALIZED_KEY_LENGTH: usize = MAX_KEY_MATERIAL_LENGTH + 4;
+
 pub const MAX_USER_ATTRIBUTE_LENGTH: usize = 256;
 
 pub const USER_ATTRIBUTE_NUMBER: u8 = 37;
--- a/src/types.rs
+++ b/src/types.rs
@@ -578,6 +578,7 @@ pub type MediumData = Bytes<MAX_MEDIUM_DATA_LENGTH>;
 pub type ShortData = Bytes<MAX_SHORT_DATA_LENGTH>;
 
 pub type Message = Bytes<MAX_MESSAGE_LENGTH>;
+pub type SerializedKey = Bytes<MAX_KEY_MATERIAL_LENGTH>;
 
 #[derive(Copy, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
 pub enum KeySerialization {
--- a/src/types.rs
+++ b/src/types.rs
@@ -568,6 +568,9 @@ pub enum Mechanism {
     X255,
     /// Used to serialize the output of a diffie-hellman
     SharedSecret,
+    Rsa2048Pkcs1v15,
+    Rsa3072Pkcs1v15,
+    Rsa4096Pkcs1v15,
 }
 
 pub type LongData = Bytes<MAX_LONG_DATA_LENGTH>;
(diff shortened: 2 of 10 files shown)


robin-nitrokey:
I think we also need some changes to the constants in `config`, right?

https://github.com/trussed-dev/trussed/compare/main...Nitrokey:trussed:main#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecd

sosthene-nitrokey:
Yes probably. I'm making the migration to a custom `Backend` in parallel and I will keep this as draft until I'm sure everything is enough.

sosthene-nitrokey:
This branch works with https://github.com/Nitrokey/trussed-rsa-backend

sosthene-nitrokey:
Looks like the CI has some issues downloading stuff to setup the docker image

sosthene-nitrokey:
I don't really understand why the CI fails

daringer:
Yeah that's weird, especially those are all urls within Azure that fail, maybe some patience will help here. On a first glance this doesn't look like something we can debug...

sosthene-nitrokey:
There are similar errors with #71

daringer:
yup yup, good indication that we should give it like 24h or something

sosthene-nitrokey:
Looks like the CI is fixed

src/config.rs
```diff
@@ -44,7 +40,16 @@ cfg_if::cfg_if! {
     }
 }
 pub const MAX_SHORT_DATA_LENGTH: usize = 128;
-pub const MAX_SIGNATURE_LENGTH: usize = 72;
+
+pub const MAX_SIGNATURE_LENGTH: usize = 512 * 2;
+// FIXME: Value from https://stackoverflow.com/questions/5403808/private-key-length-bytes for Rsa2048 Private key
+pub const MAX_KEY_MATERIAL_LENGTH: usize = 1160 * 2 + 72;
+// This is due to the fact that KEY_MATERIAL_LENGTH is bigger than MESSAGE_LENGTH for RSA.
+pub const MAX_MESSAGE_LENGTH: usize = 1024;
```
robin-nitrokey, line 48: The comment implies that the value changed, but we still use 1024.  Which one is correct?

sosthene-nitrokey: This was a remnant of the original PRs for RSA. I removed it.

src/config.rs
```diff
@@ -44,7 +41,14 @@ cfg_if::cfg_if! {
     }
 }
 pub const MAX_SHORT_DATA_LENGTH: usize = 128;
-pub const MAX_SIGNATURE_LENGTH: usize = 72;
+
+pub const MAX_SIGNATURE_LENGTH: usize = 512 * 2;
+// FIXME: Value from https://stackoverflow.com/questions/5403808/private-key-length-bytes for Rsa2048 Private key
+pub const MAX_KEY_MATERIAL_LENGTH: usize = 1160 * 2 + 72;
```
nickray, line 47: Could this be features flagged behind a `big-keys` feature perhaps? Although I'm not sure if the request enum is "big" anyway due to other variants. The thinking is that for non-RSA builds, keys are more like 32 bytes than 2-3KB.

src/types.rs
```diff
@@ -584,6 +588,12 @@ pub enum KeySerialization {
     EcdhEsHkdf256,
     Raw,
     Sec1,
+    /// Used by backends implementing RSA.
+    ///
+    /// Since RSA keys have multiple parts, and that the [`SerializeKey`](crate::api::Reply::SerializeKey) and
+    /// [`UnsafeInjectKey`](crate::api::Request::UnsafeInjectKey) have only transfer one byte array, the RSA key is serialized with postcard
+    RsaParts,
```
nickray, line 595: How about calling this `Serde` or `PostcardSerde` to make it potentially reusable?

src/config.rs
sosthene-nitrokey: The total increase to the interchange size is only 100 bytes. I'll put it behind a feature-flag.

src/types.rs
sosthene-nitrokey: I want to keep it generic and let the backend decide on the actual data format (for now at least)

src/config.rs
nickray: Yeah I agree 100 bytes is not worth the effort.

src/types.rs
sosthene-nitrokey: The serialization format doesn't affect how they're stored. I don't want to reuse `Raw` as it feels misleading. Unlike ECC, there is no standard "raw" byte representation of RSA keys. 

The "Parts" is in contrast to the PKCS format. The PKCS format is defined, but the "Parts" formats contains each element of a public key separately. The "part" name is more representative than "Internal" or "Private". "Serde" feels out of place. The use of postcard and serde is more of a workaround that the actual goal of the API. If you look at the traits in `trussed-rsa-backend`, they provide APIs directly over the structures, and hide away the parsing/serializing with postcard and serde (when possible):

```rust
fn unsafe_inject_rsa4096<'c>(
    &'c mut self,
    key_parts: RsaImportFormat,
    attributes: StorageAttributes,
) -> ClientResult<'c, reply::UnsafeInjectKey, Self>;

fn deserialize_rsa4096_public_key<'c>(
    &'c mut self,
    key_parts: RsaPublicParts,
    attributes: StorageAttributes,
) -> ClientResult<'c, reply::DeserializeKey, Self>;

```

Ideally I would even merge the `RsaImportFormat` an `RsaPublicParts` structures into Trussed proper but I want to be sure we can have something that fits all back-ends and use-cases first.

Maybe we should have mechanism extensions? That would mean more generics though.

nickray: Yeah I understand. My point is just that `Pkcs8Der` is an obviously valid general purpose format to upstream, while `RsaParts` is not. So can I'd like to merge this without this `RsaParts`. I believe you can use `Raw` (or we add something generic but general-purpose like `Experimental`) for your current work, and once it's set in stone, do a follow-up PR?

nickray: Or what about `Parts` :) That way you have your specific format to use, but unrelated algorithms might find a use too.

sosthene-nitrokey: > My point is just that `Pkcs8Der` is an obviously valid general purpose format to upstream, while `RsaParts` is not

I disagree. Whatever we do we **will** need a way to access specific parts of RSA keys. Both PIV and OpenPGP need that (OpenPGP needs it for both public and private keys, PIV at least for public keys). `Pkcs8Der` will also be incompatible with the SE050.

> Or what about `Parts` :) That way you have your specific format to use, but unrelated algorithms might find a use too.

I'm not a fan of loosing readability for more flexibility when there are no obvious benefits to it. Do you know of any other algorithm that would use this?

PR #91, opened 2023-02-06
virt: adapt to trussed-dev/littlefs2#24

daringer:
see: https://github.com/trussed-dev/littlefs2/pull/24

This is the commit to make virtual platforms work with the above PR - e.g. inside Nitrokey/nitrokey-3-firmware


daringer:
wrong repo :/

PR #93, opened 2023-02-07
Add assertions for interchange size

robin-nitrokey:
This patch adds assertions on the size of the Request, Reply and TrussedInterchange types so that we don’t accidentally increase it.

----

Increasing the interchange size can easily cost us lots of memory if many clients are used.  This patch is intended to help us keep track of the memory requirements.  Thoughts?

### DIFF
--- a/src/pipe.rs
+++ b/src/pipe.rs
@@ -81,3 +81,36 @@ pub struct ServiceEndpoint<I: 'static, C> {
 }
 
 // pub type ClientEndpoint = Requester<TrussedInterchange>;
+
+#[cfg(test)]
+mod tests {
+    use super::TrussedInterchange;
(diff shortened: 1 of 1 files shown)


sosthene-nitrokey:
I'm a bit worried of having it be always there. The size could change for reasons outside of our control like a change in the compiler or using trussed with a new target architecture. How much margin does the current assertion gives us?

Maybe these assertions could be behind a feature flag or a `--cfg`?

robin-nitrokey:
Moved into the tests.  A downside is that the tests are only executed for the x86_64 target but there should not be major differences between the targets.

sosthene-nitrokey (approved):

sosthene-nitrokey (approved):

PR #98, opened 2023-02-10
Add derive macros for Dispatch, ExtensionDispatch

robin-nitrokey:
Please don’t look at the macro implementation yet, it’s quite messy at the moment.  Instead look at the diff for `tests/backend.rs` and `tests/serde_extensions.rs` to see the derive macros in action.  The current syntax is:

```rust
#[derive(Default, trussed::backend::Dispatch)]
#[dispatch(backend_id = "Backend")]
struct Dispatch {
    test: TestBackend,
}

#[derive(Default, ExtensionDispatch)]
#[dispatch(backend_id = "Backend", extension_id = "Extension")]
#[extensions(Test = "TestExtension", Sample = "SampleExtension")]  // maps the extension ID to the type
pub struct Backends {
    #[extensions("Test")]  // declares the extensions supported by this backend
    test: TestBackend,
    #[extensions("Test", "Sample")]
    sample: SampleBackend,
}
```

What do you think?

----

Missing features:
- [ ] providing different sets of extensions for the same backend with the different IDs
- [x] having multiple `extensions` attributes to simplify conditional compilation
- [ ] warn if an extension is declared but not implemented

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,26 +34,26 @@ jobs:
           release: "9-2020-q2"
 
       - name: Build
-        run: cargo build --verbose --target ${{ matrix.target }}
+        run: cargo build --workspace --target ${{ matrix.target }}
 
       - name: Check all targets without default features
-        run: cargo check --all-targets --no-default-features
+        run: cargo check --workspace --all-targets --no-default-features
         if: matrix.target == 'x86_64-unknown-linux-gnu'
 
       - name: Check all targets with default features
-        run: cargo check --all-targets
+        run: cargo check --workspace --all-targets
         if: matrix.target == 'x86_64-unknown-linux-gnu'
 
       - name: Check all features and targets
-        run: cargo check --all-features --all-targets
+        run: cargo check --workspace --all-features --all-targets
         if: matrix.target == 'x86_64-unknown-linux-gnu'
 
       - name: Run tests
-        run: cargo test --verbose --features serde-extensions,virt
+        run: cargo test --features serde-extensions,virt
         if: matrix.target == 'x86_64-unknown-linux-gnu'
 
       - name: Check formatting
-        run: cargo fmt -- --check
+        run: cargo fmt --all -- --check
         if: matrix.target == 'x86_64-unknown-linux-gnu'
 
       - name: Check clippy lints
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,13 +1,23 @@
+[workspace]
+members = ["derive"]
+
+[workspace.package]
+authors = ["Nicolas Stalder <n@stalder.io>", "Nitrokey GmbH"]
+edition = "2021"
+homepage = "https://trussed.dev"
+license = "Apache-2.0 OR MIT"
+
 [package]
 name = "trussed"
 version = "0.1.0"
-authors = ["Nicolas Stalder <n@stalder.io>"]
-edition = "2021"
-homepage = "https://trussed.dev"
 repository = "https://github.com/trussed-dev/trussed"
-license = "Apache-2.0 OR MIT"
 description = "Modern Cryptographic Firmware"
 readme = "README.md"
+
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
 
 [dependencies]
 # general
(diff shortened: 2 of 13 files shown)


robin-nitrokey:
@daringer @sosthene-nitrokey Any thoughts on the syntax?  I think it would make it much easier to maintain the dispatch implementations.

daringer:
This hides a lot of redundant code, I love it! especially here: https://github.com/trussed-dev/trussed/pull/98/files#diff-203bd1efad2e5b766e73f7058bc930bc35004df65aeb5088a6c5845f662bdd43L383 lot of boilerplate stuff just gone... +1

sosthene-nitrokey:
My main concern is whether it can do what we do with the `Manage` extensions, that have a specific Backend ID to be used, so that an app with access to the `StagingBackend` or the `Se050Backend` cannot access it (and instead needs the `StagingManage` and `Se050Manage` backend Ids configured.

Other than that, i would put the `extensions` into the `dispatch` attribute so that only one "attribute namespace" is used.

```
#[derive(Default, trussed::backend::Dispatch)]
#[dispatch(backend_id = "Backend")]
struct Dispatch {
    test: TestBackend,
}

#[derive(Default, ExtensionDispatch)]
#[dispatch(backend_id = "Backend", extension_id = "Extension")]
#[dispatch(extensions(Test = "TestExtension", Sample = "SampleExtension"))]  // maps the extension ID to the type
pub struct Backends {
    #[dispatch(extensions("Test"))]  // declares the extensions supported by this backend
    test: TestBackend,
    #[dispatch(extensions("Test", "Sample"))]
    sample: SampleBackend,
}
```

robin-nitrokey:
> My main concern is whether it can do what we do with the `Manage` extensions, that have a specific Backend ID to be used, so that an app with access to the `StagingBackend` or the `Se050Backend` cannot access it

I don’t see a problem with that.  We can just use fields with different names but the same type.  I think something like this should already work with the current implementation:

```rust
#[derive(Default, ExtensionDispatch)]
#[dispatch(backend_id = "Backend", extension_id = "Extension")]
#[extensions(Test = "TestExtension", Special = "SpecialExtension", Sample = "SampleExtension")]  // maps the extension ID to the type
pub struct Backends {
    #[extensions("Test")]  // declares the extensions supported by this backend
    test: TestBackend,
    #[extensions("Test", "Special")]
    test_special: TestBackend,
    #[extensions("Test", "Sample")]
    sample: SampleBackend,
}
```

> Other than that, i would put the `extensions` into the `dispatch` attribute so that only one "attribute namespace" is used.

I also thought about that, but on the other hand, it just adds more boilerplate code without real benefit.  And looking at the most popular derive crates, using multiple attributes is quite common.  For example, `clap` uses multiple attributes to group options logically.

robin-nitrokey:
If we don’t want to duplicate the backend, we could probably use a special attribute like:

```rust
#[derive(Default, ExtensionDispatch)]
#[dispatch(backend_id = "Backend", extension_id = "Extension")]
#[extensions(Test = "TestExtension", Special = "SpecialExtension", Sample = "SampleExtension")]
pub struct Backends {
    #[extensions("Test")]
    test: TestBackend,

    #[extensions("Test", "Special")]
    #[dispatch(delegate_to = "test")]
    test_special: TestBackend,

    #[extensions("Test", "Sample")]
    sample: SampleBackend,
}
```

sosthene-nitrokey:
> If we don’t want to duplicate the backend, we could probably use a special attribute like:
> 
> ```rust
> #[derive(Default, ExtensionDispatch)]
> #[dispatch(backend_id = "Backend", extension_id = "Extension")]
> #[extensions(Test = "TestExtension", Special = "SpecialExtension", Sample = "SampleExtension")]
> pub struct Backends {
>     #[extensions("Test")]
>     test: TestBackend,
> 
>     #[extensions("Test", "Special")]
>     #[dispatch(delegate_to = "test")]
>     test_special: TestBackend,
> 
>     #[extensions("Test", "Sample")]
>     sample: SampleBackend,
> }
> ```

Would that work for a `#[derive]` macro? I think `test_special` would still be a member of the struct at the end since derive macros can't change the struct definition itself.

robin-nitrokey:
Hmm, then we could change the type to `()`.  Anyway, I would suggest to first merge a version without duplicate backends.  We can then add that feature in a second step.

sosthene-nitrokey:
Yes

robin-nitrokey:
Updated:
- Rebased onto main.
- Upated `syn` dependency to v2.
- Made the implementation more robust and improved error messages.
- Added examples.

robin-nitrokey:
Updated:
- Made the `extensions` attribute optional for backends in `ExtensionDispatch`.
- Added an `ExtensionId` derive macro that implements `From<T> for u8` and `TryFrom<u8> for T`.

sosthene-nitrokey (approved):

PR #99, opened 2023-02-10
Migrate to interchange 0.3.0

sosthene-nitrokey:
This PR migrates to interchange 0.3.0, using const generics.

I'm also looking to see if we can benefit from the improved flexibility to get rid of the `client-N` feature by having the interchange be provided by the runner instead of coming from trussed.


robin-nitrokey:
I had a similar patch for the old client context implementation.  It replaced the `client-?` features with const generics and a lifetime in `Service` and `ServiceEndpoint`:

https://github.com/robin-nitrokey/trussed/commit/bd18de095cb37d7a686c257823c3bae93a960b69

robin-nitrokey (approved):
Thanks, LGTM!  Maybe squash before merging.

PR #100, opened 2023-02-13
Make the interchange be provided by the runner

sosthene-nitrokey:
This allows for more flexibility. Ultimately the goal it to enable running multiple trussed instances at the same time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ once_cell = "1.13.0"
 # rand_core = { version = "0.5", features = ["getrandom"] }
 
 [features]
-default = ["default-mechanisms", "clients-5"]
+default = ["default-mechanisms"]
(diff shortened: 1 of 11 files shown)


robin-nitrokey:
Please don’t remove the `StoreProvier::ifs` function.  It is needed to setup the provisioner app in our usbip runner.  https://github.com/Nitrokey/nitrokey-3-firmware/blob/5ecad2efb1fc07dfc27b24686ce36d49dd59976c/runners/usbip/src/main.rs#L196

sosthene-nitrokey:
I think there's a better solution that doesn't require lifetimes everywhere. If we make `Service::process` take as input an array of `Responder`, it doesn't need to own them any more. There would be some more lifetimes in `process` but those would be elided very easily.

PR #101, opened 2023-02-14
Add raw RSA mechanism

sosthene-nitrokey:
This mechanism would be required for PIV.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -564,6 +564,20 @@ pub enum Mechanism {
     X255,
     /// Used to serialize the output of a diffie-hellman
     SharedSecret,
+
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -564,6 +564,17 @@ pub enum Mechanism {
     X255,
     /// Used to serialize the output of a diffie-hellman
     SharedSecret,
+
+    /// Exposes the Raw RSA encryption/decryption primitive. Be aware this is dangerous.
```
robin-nitrokey, line 568: We should also briefly explain *why* it is dangerous.

sosthene-nitrokey: Done

robin-nitrokey (approved):

PR #103, opened 2023-03-01
Remove firmware compilation tests

sosthene-nitrokey:
This will allow breaking change for an 0.2 release

### DIFF
--- a/.github/workflows/firmwares.yml
+++ b/.github/workflows/firmwares.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Firmwares
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


robin-nitrokey:
Can’t we just remove/comment the `on` triggers instead of commenting out all steps?

sosthene-nitrokey:
There's actually an even better solution. We can manually disable it in the GH UI it seems: https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow. I think this is what we want. The two workflows were both named "CI" so I renamed it to "Firmwares". I uncommented the Nitrokey firmware too, so that we can just keep the file as-is and disable the workflow.

sosthene-nitrokey:
Should we merge and disable now or wait for Friday's steering call?

daringer:
let's wait for the steering call to align on a path how (and when) to migrate solo2 towards the breaking changes

PR #104, opened 2023-03-09
Allow iterating over the external and volatile filesystems

sosthene-nitrokey:
This PR fixes a previous limitation where the methods for iterating over the filesystem only allowed iterating over the internal filesystem.

PR #96, opened 2023-02-08
Bump littlefs2 to 0.4.0

robin-nitrokey:
To patch:
- [ ] fido-authenticator: https://github.com/solokeys/fido-authenticator/pull/23
- [ ] lpc55-hal: https://github.com/lpc55/lpc55-hal/pull/60

PR #107, opened 2023-03-14
Iteration optimisation

sosthene-nitrokey:
This is built on top of #104 and #96, and https://github.com/trussed-dev/littlefs2/pull/33

This PR:

Uses the apis added to https://github.com/trussed-dev/littlefs2/pull/33 to improve the skipping that happens each time in `read_dir(_files)_next()`. I think the operation is likely still $O(n)$ (with n being the number of files in the directory, but it should be much faster.

It also exposes `read_dir(_files)_nth` syscalls that make it possible to resume iteration at a certain point without having to iterate n time with `read_dir(_files)_next`, which is a $O(n^2)$ operation


sosthene-nitrokey:
The bug mentioned in https://github.com/trussed-dev/trussed/pull/107 does not affect this PR anyway because the current behaviour of iteration is to consider any filesystem error as the end of the iteration, which means the bug does not change observed behaviour.

nickray:
In favor of this, needs some conflict fixing.

PR #111, opened 2023-03-30
Put debug-dump-store behind a feature-flag

sosthene-nitrokey:
When debugging https://github.com/Nitrokey/piv-authenticator/pull/12 I realized this was not removed from production firmware. Putting it behind a feature flag will ensure that it does.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -312,6 +312,9 @@ impl<P: Platform> ServiceResources<P> {
                 info_now!(":: PERSISTENT");
                 recursively_list(self.platform.store().ifs(), path!("/"));
 
+                info_now!(":: EXTERNAL");
+                recursively_list(self.platform.store().efs(), PathBuf::from("/"));
+
                 info_now!(":: VOLATILE");
                 recursively_list(self.platform.store().vfs(), path!("/"));
 
--- a/src/service.rs
+++ b/src/service.rs
@@ -307,6 +307,7 @@ impl<P: Platform> ServiceResources<P> {
 
             // This is now preferably done using littlefs-fuse (when device is not yet locked),
             // and should be removed from firmware completely
+            #[cfg(debug_dump_store)]
             Request::DebugDumpStore(_request) => {
 
                 info_now!(":: PERSISTENT");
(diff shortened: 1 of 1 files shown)


robin-nitrokey (approved):

daringer (approved):

szszszsz:
I thought the consensus is to not make a feature out of the debug switches. Has this changed?

robin-nitrokey:
I think this is a misunderstanding.  We almost never use rustc’s `--cfg` flag directly, just features.  But in the source code, we use the `cfg` attribute to check for that feature.

sosthene-nitrokey:
I think the main difference between the two is that the addition of the debug-dump-store is purely additive, whereas the removal of encryption is not.

This flag does not change the test paths if you use `--all-features`, since it only adds new code.
For the encryption on the other hand, it removes code, so testing with `--all-features` does not test encryption.

Also removal of encryption is not something we want to encourage at any point, while features are easily discoverable, and therefore can mistakenly be enabled, which would be highly problematic.

sosthene-nitrokey:
It's also consistent with the `std` features, that are here only for debugging/testing. They can be features because they're additive, and not *that dangerous*, contrary to the encryption flag.

szszszsz:
`dump-store` in production firmware sounds dangerous to me, hence the question

sosthene-nitrokey:
It makes the call available but you would still have to call it, and you would have to have the logs somehow accessible on a production device before this becomes really dangerous.

nickray:
Did you come to consensus? :) Slight tendency to cfg-flag on my side (default "off" naturally).

sosthene-nitrokey:
I moved it behind a CFG. This doesn't matter much if you prefer cfg we can merge it that way.

robin-nitrokey:
@sosthene-nitrokey What do you think about removing the syscall implementation entirely and always returning `RequestNotAvailable`?  The current implementation does not do anything anyway and it is not actually used.  Then we could drop the syscall in the next breaking `trussed-core` release.

daringer:
guess we can close this in favor of #205

robin-nitrokey:
Replaced by https://github.com/trussed-dev/trussed/pull/205.

PR #115, opened 2023-04-11
Add wrap_key_to_file syscall

sosthene-nitrokey:
These syscall are required because RSA keys can get bigger than the data that can be transmitted through `wrap_key`.


robin-nitrokey:
I think we can close this for the time being and create a new PR if we want to upstream the extension.

PR #116, opened 2023-04-14
Chunk api

sosthene-nitrokey:
This API adds a streaming API to write large files.

It stores chunks in volatile memory, and then commits them to static memory when writing is over.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -50,7 +50,10 @@ generate_enums! {
     SerializeKey: 17
     Sign: 18
     WriteFile: 19
-    WriteChunk: 66
+    StartChunkedWrite: 66
+    WriteChunk: 67
+    FlushChunks: 68
+    AbortChunkedWrite: 69
     UnsafeInjectKey: 20
     UnsafeInjectSharedKey: 21
     UnwrapKey: 22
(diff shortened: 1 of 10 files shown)


sosthene-nitrokey:
We might also want to consider an AEAD version of this API to be able to encrypt large files, likely using [aead::stream](https://docs.rs/aead/latest/aead/stream/index.html)

sosthene-nitrokey:
I'm making this a draft. While implementing the chunked encrypted writes I'm realizing this API will need some modifications.

robin-nitrokey:
Re. https://github.com/trussed-dev/trussed/pull/115 and this PR:  We talked about using extensions as a tool for experimenting with new requests before they are stabilized and merged into upstream Trussed.  Would that make sense here?

sosthene-nitrokey:
I agree. I mentioned that in https://github.com/Nitrokey/trussed/pull/23

I think it would make sense to have these as extensions. The Chunk API could even use allocation in that case, and be much more straightforward for encrypted chunked writes (we could store everything in a `alloc::Vec<u8>` and encrypt/decrypt it with an AEAD in one go).

sosthene-nitrokey:
Making it an extension would still require updates to the `filestore` trait and implementation.

robin-nitrokey:
Isn’t it sufficient to access the `Store` directly?

sosthene-nitrokey:
That's also an option

robin-nitrokey:
I think we can close this for the time being and create a new PR if we want to upstream the extension.

PR #119, opened 2023-04-23
Add helpers for backend stores

robin-nitrokey:
This patch adds helpers for accessing the global and client backend stores.


robin-nitrokey:
We should probably use different types for the stores to avoid confusion.

nickray:
This doesn't compile after rebasing on `origin/main`.

PR #122, opened 2023-04-27
Use a reference to the interchange data instead of copying it to the stack

sosthene-nitrokey:
In theory this should reduce memory usage and improve performance.

On hardware I did not notice an actual performance improvement though.


sosthene-nitrokey:
Checking with `-Z emit-stack-sizes`: this makes the trussed task go from 21784 to 16976 bytes of stack usage.

szszszsz (approved):

daringer (approved):

PR #123, opened 2023-05-09
filestore: Add remove_dir_all_where method

sosthene-nitrokey:
This will be useful for extensions, even if it's not used in core trussed


daringer (approved):

daringer:
on top it simply wraps a `store::` primitive
so: lgtm

PR #125, opened 2023-05-15
Support cancellation of requests

sosthene-nitrokey:
See https://github.com/trussed-dev/trussed/discussions/124

### DIFF
--- a/src/interrupt.rs
+++ b/src/interrupt.rs
@@ -1,0 +1,74 @@
+use core::{
+    fmt::Debug,
+    sync::atomic::{AtomicU8, Ordering::Relaxed},
+};
+
+#[derive(Default, Debug, PartialEq, Eq)]
+pub enum InterruptState {
+    #[default]
+    Idle = 0,
+    Working = 1,
+    Interrupted = 2,
+}
+
+impl TryFrom<u8> for InterruptState {
+    type Error = ();
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(Self::Idle),
+            1 => Ok(Self::Working),
+            2 => Ok(Self::Interrupted),
+            _ => Err(()),
+        }
+    }
+}
+
+impl From<InterruptState> for u8 {
+    fn from(value: InterruptState) -> Self {
+        value as _
+    }
+}
+
+#[derive(Default)]
+pub struct InterruptFlag(AtomicU8);
+
+const CONV_ERROR: &str =
+    "Internal trussed error: InterruptState must always be set to an enum variant";
+
+impl InterruptFlag {
+    pub const fn new() -> Self {
+        Self(AtomicU8::new(0))
+    }
+    fn load(&self) -> InterruptState {
+        self.0.load(Relaxed).try_into().expect(CONV_ERROR)
+    }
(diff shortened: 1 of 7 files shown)


src/service.rs
```diff
@@ -813,7 +813,15 @@ impl<P: Platform, D: Dispatch> Service<P, D> {
...
-                ep.interchange.respond(reply_result).ok();
+                match ep.interchange.respond(reply_result) {
+                    Ok(()) => {}
+                    Err(_) => {
+                        info!("Cancelled request");
+                        if ep.interchange.is_canceled() {
+                            ep.interchange.acknowledge_cancel().ok();
+                        }
+                    }
+                };
```
robin-nitrokey, line 837: nit: `if let`?

src/interrupt.rs
```diff
@@ -0,0 +1,74 @@
...
+#[derive(Default, Debug, PartialEq, Eq)]
+pub enum InterruptState {
+    #[default]
+    Idle = 0,
+    Working = 1,
+    Interrupted = 2,
+}
+
+impl TryFrom<u8> for InterruptState {
+    type Error = ();
```
robin-nitrokey, line 15: nit: use proper error type

src/service.rs
```diff
@@ -523,6 +523,12 @@ impl<P: Platform> ServiceResources<P> {
                 let previous_status = self.platform.user_interface().status();
                 self.platform.user_interface().set_status(ui::Status::WaitingForUserPresence);
                 loop {
+                    if ctx.interrupt.map(|i|i.is_interrupted()) == Some(true) {
+                        // Error does not matter as it will be dropped anyway
+                        info_now!("User presence request cancelled");
+                        return Ok(reply::RequestUserConsent{result: Err(consent::Error::Interrupted)}.into());
+                    }
```
robin-nitrokey, line 530: nit: formatting is off here (missing spaces)

robin-nitrokey (approved):
LGTM!

src/service.rs
sosthene-nitrokey: Even better:

```rust
                if ep.interchange.respond(reply_result).is_err() && ep.interchange.is_canceled() {
                    info!("Cancelled request");
                    ep.interchange.acknowledge_cancel().ok();
                };
```

src/service.rs
sosthene-nitrokey: Weird, rustfmt does not format this part.

running with ` --config error_on_unformatted=true` gives

```
rustfmt --edition 2018 src/service.rs --config error_on_unformatted=true
error[internal]: not formatted because a comment would be lost
   --> /home/soso-nitrokey/Code/trussed/src/service.rs:149
    |
149 |         match request {
    |
    = note: set `error_on_unformatted = false` to suppress the warning against comments or string literals

warning: rustfmt has failed to format. See previous 1 errors.
```

I'll manually fix this formatting, but if we remove the offending comment, it changes the entire `match` statement, so it would make a large diff.

It should probably be done in another PR.

src/interrupt.rs
sosthene-nitrokey: Done

robin-nitrokey (approved):

PR #130, opened 2023-07-19
Remove restoration of ui status after RequestUserPresence call

### LINKED ISSUE #129: RequestUserConsent does not properly restore the ui status when interrupted or timed out.
In the [RequestUserConsent](https://github.com/trussed-dev/trussed/blob/main/src/service.rs#L567) syscall, the ui status is saved, then set to `WaitingForUserPresence` and finally restored at the end of the call.

However, there are early returns in the loop that cause the restoration to not happen. This is not a very large issue as the call to `process` [restores the status to `Idle`](https://github.com/trussed-dev/trussed/blob/main/src/service.rs#L843).
I think the restoration of the status should be removed as it isn't very useful since it is overridden to Idle just after.

sosthene-nitrokey:
The `reply_to` method always restores the status to Idle anyway after the syscall returns

Close #129

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -523,7 +523,6 @@ impl<P: Platform> ServiceResources<P> {
                 let timeout =
                     core::time::Duration::from_millis(request.timeout_milliseconds as u64);
 
-                let previous_status = self.platform.user_interface().status();
                 self.platform
                     .user_interface()
                     .set_status(ui::Status::WaitingForUserPresence);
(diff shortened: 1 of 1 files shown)


nickray:
Restoring the status in all cases could work by wrapping the `loop` in a closure, so we can get the result, always restore status, and only then return.

sosthene-nitrokey:
It would still be overridden by the `reply_to` method anyway.

PR #135, opened 2023-10-25
Remove unnecessary request copy

robin-nitrokey:
This patch removes an unnecessary take_request call from Service::process, reducing the stack usage by roughly 5 kB.


robin-nitrokey:
Already reviewed an approved in https://github.com/Nitrokey/trussed/pull/30.

sosthene-nitrokey (approved):

PR #141, opened 2023-11-13
Lazy stores

### LINKED ISSUE #136: Performance of `ServiceResources` in request handling
When an app has backends configured that deal with `core` requests, they will "see" each core API request.

But most backends get all the "resources" from `ServiceResources`, but loading the keystore necessitates forking the RNG (because the keystore needs an RNG for `KeyId` generation. The same is true for  the `cerstore` and the `counterstore`.

However many requests will never touch any of these stores and those that do will probably only ever touch one.

Since generating RNG is expensive (on embedded harware), if each intermediary backend that sees the requests forks the RNG that way it might add up to a significant overhead.
For example in opcard, which has the RSA backend enabled, each requests to the core API ends up forking the RNG at least 5 times:

- https://github.com/Nitrokey/trussed-rsa-backend/blob/main/src/lib.rs#L441
- https://github.com/Nitrokey/trussed-rsa-backend/blob/main/src/lib.rs#L442
- https://github.com/trussed-dev/trussed/blob/main/src/service.rs#L144
- https://github.com/trussed-dev/trussed/blob/main/src/service.rs#L145
- https://github.com/trussed-dev/trussed/blob/main/src/service.rs#L146

And then [some requests](https://github.com/trussed-dev/trussed/blob/main/src/service.rs#L146) fork it once more through `self.rng`.

This seems unnecessary and could be one of the reason of the slowness of opcard.

I see two way to reduce the number of forking of the RNG:

- Avoid using `self.rng()` (I'm guilty of this in backends) and instead use the already-forked RNG from the `keystore` [like the crypto mechanisms do](https://github.com/trussed-dev/trussed/blob/main/src/mechanisms/x255.rs#L89)
- Use lazily intialize the `keystore`, `certstore` and `counterstore` only for requests that need it.

Ideally we could also totally avoid forking the RNG for requests that only need to read data from the keystore and don't need to generate `KeyId`s.

sosthene-nitrokey:
Built on top of #140  and #139 

Make the store loading lazy, therefore reducing the need to fork the trussed RNG and improving performance

Fix #136

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -139,11 +139,19 @@ impl<P: Platform> ServiceResources<P> {
         // TODO: what we want to do here is map an enum to a generic type
         // Is there a nicer way to do this?
 
+        /// Coerce an FnMut into a FnOnce to ensure the stores are not created twice by mistake
+        fn once<R, P>(
+            generator: impl FnMut(&mut ServiceResources<P>, &mut CoreContext) -> R,
+        ) -> impl FnOnce(&mut ServiceResources<P>, &mut CoreContext) -> R {
+            generator
+        }
+
         let full_store = self.platform.store();
 
-        let keystore = &mut self.keystore(ctx.path.clone())?;
-        let certstore = &mut self.certstore(ctx)?;
-        let counterstore = &mut self.counterstore(ctx)?;
+        let keystore = once(|this, ctx| this.keystore(ctx.path.clone()));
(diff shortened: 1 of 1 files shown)


daringer:
the `once` implementation is really fancy to achieve this, lgtm

robin-nitrokey (approved):

PR #142, opened 2023-11-22
Add rename syscall to FilesystemClient

robin-nitrokey:
This patch adds a rename syscall to the FilesystemClient that allows applications to move a file within the same storage location and within the client namespace.


sosthene-nitrokey:
Would we not want to make it possible to rename across filesystems ?

It's a bit more involved but it's already implemented in trussed-staging as an internal need of the staging backend: https://github.com/Nitrokey/trussed-staging/blob/c6aa6bdd65f04eb746a2d65725ba2d01c27ae1f7/src/streaming/store.rs#L99

robin-nitrokey:
To me it makes sense to have `rename` as a cheap and atomic operation that is only possible on the same filesystem.  Adding the method you linked would be nice, but I’d rather call it `copy` to reflect the actual cost.

src/client.rs
```diff
@@ -654,6 +654,19 @@ pub trait FilesystemClient: PollClient {
         self.request(request::Metadata { location, path })
     }
 
+    /// Rename a file or directory.
+    ///
+    /// If `to` exists, it must be the same type as `from`.  If `to` is a directory, it must be
```
sosthene-nitrokey, line 659: nit: Outside of the littlefs documentation, `type` does not feel to have an obvious meaning.

```suggestion
    /// If `to` exists, it must be the same type (file or directory) as `from`.  If `to` is a directory, it must be
```

sosthene-nitrokey (approved):

src/client.rs
robin-nitrokey: Fixed.

PR #143, opened 2023-12-12
Put unused syscalls behind features

robin-nitrokey:
CounterClient and CryptoClient::attest are not currently used by the
solo2 or nk3 firmwares, but due to the indirect dispatch of Trussed
requests, they cannot be optimized out by the linker.  This patch
introduces feature flags for these syscalls that are enabled by default.


sosthene-nitrokey:
Tests need the feature flags too

sosthene-nitrokey (approved):

daringer (approved):

robin-nitrokey:
Argh, I think adding it to the default features is a problem for us:  We would have to modify every dependency using Trussed to *not* select the default features.  I don’t see a good way to solve this in upstream.  Should we use a fork for the moment?

sosthene-nitrokey:
I don't see a better solution

PR #144, opened 2023-12-15
Use DynFilesystem trait for FS operations

robin-nitrokey:
This reduces code complexity and binary size.

### DIFF
--- a/src/store/filestore.rs
+++ b/src/store/filestore.rs
@@ -1,8 +1,8 @@
 use crate::{
     error::{Error, Result},
     // service::ReadDirState,
-    store::{self, Store},
-    types::{LfsStorage, Location, Message, UserAttribute},
+    store::{self, DynFilesystem, Store},
+    types::{Location, Message, UserAttribute},
     Bytes,
 };
 use littlefs2::path;
@@ -27,7 +27,6 @@ use littlefs2::{
     path::{Path, PathBuf},
 };
 
-use super::Fs;
 pub type ClientId = PathBuf;
 
 pub struct ClientFilestore<S>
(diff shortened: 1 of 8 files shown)


src/store/filestore.rs
```diff
@@ -197,21 +196,22 @@ impl<S: Store> ClientFilestore<S> {
             })
             .ok())
     }
-    fn read_dir_next_impl<F: LfsStorage + 'static>(
+    fn read_dir_next_impl(
         &mut self,
         state: ReadDirState,
-        fs: &'static Fs<F>,
     ) -> Result<Option<(DirEntry, ReadDirState)>> {
+        // TODO: check what happens (or used to happen) for FS != IFS
```
sosthene-nitrokey, line 203: Why?

src/store/filestore.rs
```diff
@@ -280,7 +280,8 @@ impl<S: Store> ClientFilestore<S> {
                             real_dir: dir.clone(),
                             last: i,
                             location,
-                            user_attribute,
+                            // TODO: check if we can avoid that clone
```
sosthene-nitrokey, line 283: The `Dyn` impl could actually take an `FnOnce`, turn it into an `FnMut` that panics if invoked more than once.

sosthene-nitrokey (approved):

src/store/filestore.rs
robin-nitrokey: I think I saw a place where this was always called with the IFS even for other locations, but I can’t find it any more.  Maybe I accidentally looked at the `locate_file` implementation.

robin-nitrokey: Removed the comment.

src/store/filestore.rs
robin-nitrokey: I think we should also be able to just move that part of the code out of the closure.

robin-nitrokey: Yeah, that works and also gets rid of the dir clone.

PR #145, opened 2023-12-15
 Remove type parameters from Store 

robin-nitrokey:
Depends on:
- https://github.com/trussed-dev/trussed/pull/144

Alternatively, we could completely drop the `Store` trait and just use a struct with three references.  To be tested whether that is more efficient.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,3 +127,6 @@ test-attestation-cert-ids = []
 [package.metadata.docs.rs]
 features = ["serde-extensions", "virt"]
 rustdoc-args = ["--cfg", "docsrs"]
+
+[patch.crates-io]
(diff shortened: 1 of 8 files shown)


sosthene-nitrokey:
Is this really necessary?

robin-nitrokey:
It is not necessary, but I think it has multiple benefits:  It leads to simpler code.  It allows us to get rid of the unnecessary `Fs` type that is just another level of indirection and another `static mut`.  And it allows us to be more flexible in `Store` implementations, especially e. g. for the virtual runners or when replacing a storage with RAM.  So I would rather ask the question:  Why do we need these type parameters?

sosthene-nitrokey:
Good!

robin-nitrokey:
Replaced by https://github.com/trussed-dev/trussed/pull/195.

PR #150, opened 2024-03-04
ClientFilestore: make the `dat` folder optional

sosthene-nitrokey:
This allows custom backends to improve their storage, reducing the number of required intermediary directories.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -113,6 +113,11 @@ impl<P: Platform> ServiceResources<P> {
         ClientFilestore::new(client_id, self.platform.store())
     }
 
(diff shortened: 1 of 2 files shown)


robin-nitrokey (approved):
Maybe it would make sense to also have this distinction in the type system in the long term, e. g. use a `RawClientFilestore` type, or `ScopedFilestore<Client>`, `ScopedFilestore<RawClient>`, ...?  I fear that the current implementation could lead to subtle bugs when the wrong constructor is used.

sosthene-nitrokey:
In the future yes. But not until we remove the configuration from the trussed-auth and se050 backends. Otherwise we can't just change the constructor based on the config.

PR #151, opened 2024-03-07
read_dir_first: stop at the first file that is alphabetically "after" `not_before`

sosthene-nitrokey:
In fido-authenticator, if we change the paths of RK  to be: "rp_id.rk_id" instead of the current "rp_id/rk_id", we still want to be able to iterate over the keys even though we only  know the "rp_id" and not the "rk_id". Therefore we need to be able to stop at "rp_id.***" when giving "rp_id" in `not_before`

This is technically a breaking change because now, given the files:

- "aaa"
- "aaabbb"

 "read_dir_first"  with "aaa" as `not_before` would only yield "aaa"
due to https://github.com/littlefs-project/littlefs/issues/923. Now this will yield both files, and yield "aaabbb" first.

I beleive this behaviour is technically more correct as it is likely what would be expected to be yield expecting alphabetical order (though the order of the entries is still incorrect).

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -244,7 +244,7 @@ pub mod request {
         ReadDirFirst:
           - location: Location
           - dir: PathBuf
-          - not_before_filename: Option<PathBuf>
+          - not_before: Option<(PathBuf, bool)>
 
         ReadDirNext:
 
--- a/src/api.rs
+++ b/src/api.rs
@@ -17,6 +17,32 @@ mod macros;
 //
 // At minimum, we don't want to list the indices (may need proc-macro)
 
+#[derive(Clone, Eq, PartialEq, Debug, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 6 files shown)


src/api.rs
```diff
@@ -244,7 +244,7 @@ pub mod request {
         ReadDirFirst:
           - location: Location
           - dir: PathBuf
-          - not_before_filename: Option<PathBuf>
+          - not_before: Option<(PathBuf, bool)>
```
robin-nitrokey, line 247: I don’t like this tuple, it’s very hard to understand.  What do you think about moving the `bool` into a separate field that only has an effect if `not_before_filename` is set?  Alternatively, a helper struct could be used.

sosthene-nitrokey: I replaced it with an enum with 3 states.

robin-nitrokey (approved):

PR #153, opened 2024-03-25
derive: Add options for ExtensionDispatch

robin-nitrokey:
This patch adds more options to the ExtensionDispatch derive macro so that it can be used for nitrokey-3-firmware:
- `delegate_to` makes it possible to create an alias for a backend using a different set of extensions.  To avoid misunderstandings, delegating backends are forced to use the unit type ().
- `skip` ignores fields.
- `no_core` skips backends when dispatching core requests.


robin-nitrokey:
ping @daringer @sosthene-nitrokey

daringer (approved):

PR #158, opened 2024-04-02
Add OptionalBackend wrapper

robin-nitrokey:
Backends can be activated at runtime depending on the configuration or execution environment.  This patch adds an OptionalBackend wrapper that implements Backend and ExtensionImpl but returns a RequestNotAvailable error if the inner backend is not set.

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -92,3 +92,60 @@ impl TryFrom<u8> for NoId {
         Err(Error::InternalError)
     }
 }
+
+/// Helper type for optional backends.
+///
+/// If the backend is `None`, [`Error::RequestNotAvailable`][] is returned.
+#[derive(Debug)]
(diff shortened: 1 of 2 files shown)


src/backend.rs
```diff
@@ -92,3 +92,50 @@ impl TryFrom<u8> for NoId {
         Err(Error::InternalError)
     }
 }
+
+/// Helper type for optional backends.
+///
+/// If the backend is `None`, [`Error::RequestNotAvailable`][] is returned.
+#[derive(Debug)]
+pub struct OptionalBackend<T>(pub Option<T>);
```
sosthene-nitrokey, line 100: nit: I would make the inner field private and add a `fn new` and `fn into_inner` instead.

sosthene-nitrokey (approved):

sosthene-nitrokey (approved):

PR #162, opened 2024-04-10
Unify mechanism traits

robin-nitrokey:
This patch combines the operation traits that were previously used to call mechanism implementations into a single MechanismImpl trait.  This has several advantages:
- We can use a macro to implement the dispatch from the Mechanism enum, removing boilerplate code from the reply_to implementation.
- To implement an operation for a mechanism, it is now sufficient to override the respective trait method.  It is no longer necessary to also update reply_to.
- The need to annotate all mechanism methods with #[inline(never)] to avoid producing a huge reply_to function (see the comment in mechanisms.rs) is reduced as we can just mark the methods generated by the macro as #[inline(never)].
- This reduces the binary size required in the stable nitrokey-3-firmware by some kB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Removed the `Copy` requirement.
   - Removed the `unsafe` keyword for the `Store` trait.
 - Removed the `unsafe` keyword for the `Platform` trait.
+- Replaced the mechanism RPC traits in `service` with a single `MechanismImpl` trait.
 
 ### Fixed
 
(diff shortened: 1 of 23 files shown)


sosthene-nitrokey:
I like that

robin-nitrokey:
Rebased onto main and updated so that we now only define a single list of implemented mechanisms.

sosthene-nitrokey (approved):

sosthene-nitrokey:
I don't really like `paste`, the alternative that I tend to use is to have the same name written multiple times for all the versions. It's a bit more noisy but more readable, and I find the less dependencies the better.

robin-nitrokey:
Unfortunately `paste` has been [marked as unmaintained](https://rustsec.org/advisories/RUSTSEC-2024-0436.html) since I created the first version of this PR, so using it would probably lead to problems sooner or later anyway.  I’ve updated the PR to remove the dependency.

sosthene-nitrokey (approved):

PR #166, opened 2024-06-07
Fix warnings

sosthene-nitrokey:


robin-nitrokey:
I think this PR is based on a wrong branch.  Only the last commit is really relevant, right?

PR #163, opened 2024-04-19
Add support for more elliptic curves

sosthene-nitrokey:
- [x] p348 (using RustCrypto)
- [x] p521 (using Ruscrypto)
- [x] brainpoolP256r1 (Mechanism definition only, implementation in separate backend)
- [x] brainpoolP384r1 (Mechanism definition only, implementation in separate backend)
- [x] brainpoolP512r1  (Mechanism definition only, implementation in separate backend)

PR #167, opened 2024-07-23
Add support for more elliptic curves

sosthene-nitrokey:
- [x] p348 (using RustCrypto)
- [x] p521 (using Ruscrypto)
- [x] brainpoolP256r1 (Mechanism definition only, implementation in separate backend)
- [x] brainpoolP384r1 (Mechanism definition only, implementation in separate backend)
- [x] brainpoolP512r1  (Mechanism definition only, implementation in separate backend)

This PR is the upstream version of https://github.com/trussed-dev/trussed/pull/163

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
         if: matrix.target == 'x86_64-unknown-linux-gnu'
 
       - name: Run tests
-        run: cargo test --features serde-extensions,virt
+        run: cargo test --features serde-extensions,virt,p384,p521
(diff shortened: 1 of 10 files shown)


robin-nitrokey (approved):
I noticed that the there are no client traits for the brainpool curves.  No big issue for me because I don’t really like the API duplication between `CryptoClient` and the mechanisms.  But is there a special reason for that?

Otherwise LGTM.

sosthene-nitrokey:
Good point.

I did not add the client traits because I don't use them in opcard, so they would not end up being used anyway.
We can add them in the future if needed.

PR #168, opened 2024-08-02
key deletion: delete first the faster volatile storage and test publicc keys first

sosthene-nitrokey:
Most key deletions are for volatile public keys (temporary keys for FIDO pin protocol, PIN keys from trussed-auth etc...). In any cases, persistent keys are more rarely deleted, and volatile is the fastest storage, so it being first is overall a performance improvement.

I think long term (once we have the builder-pattern based syscall implementation maybe?) we should add optional location and secrecy arguments to the syscall. It is rare that the caller would not know the kind of key it is deleting.

Current order is:

- Private
  - Internal
  - External
  - Volatile 
- Public 
  - Internal
  - External
  - Volatile

This PR makes it:

- Volatile
  - Public
  - Private
- External
  - Public
  - Private
- Internal
  - Public
  - Private

### DIFF
--- a/src/store/keystore.rs
+++ b/src/store/keystore.rs
@@ -138,15 +138,15 @@ impl<S: Store> Keystore for ClientKeystore<S> {
 
     // TODO: is this an Oracle?
     fn delete_key(&self, id: &KeyId) -> bool {
-        let secrecies = [key::Secrecy::Secret, key::Secrecy::Public];
+        let secrecies = [key::Secrecy::Public, key::Secrecy::Secret];
 
-        let locations = [Location::Internal, Location::External, Location::Volatile];
+        let locations = [Location::Volatile, Location::Internal, Location::External];
 
-        secrecies.iter().any(|secrecy| {
-            let path = self.key_path(*secrecy, id);
-            locations
-                .iter()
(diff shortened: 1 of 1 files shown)


src/store/keystore.rs
```diff
@@ -142,11 +142,11 @@ impl<S: Store> Keystore for ClientKeystore<S> {
 
         let locations = [Location::Internal, Location::External, Location::Volatile];
 
-        secrecies.iter().any(|secrecy| {
-            let path = self.key_path(*secrecy, id);
-            locations
-                .iter()
-                .any(|location| store::delete(self.store, *location, &path))
+        locations.iter().rev().any(|location| {
```
robin-nitrokey, line 145: nit: changing the order in `locations` and `secrecies` would be more intuitive than reversing it

robin-nitrokey (approved):
I would assume that internal is faster than external.  Why not Volatile / Internal / External?

sosthene-nitrokey:
Yeah, and apps using internal storage are also probably more time-sensitive. I'll change the order.

robin-nitrokey (approved):

daringer (approved):

Issue #177, opened 2024-10-31
Add trussed-core crate


robin-nitrokey:
- [x] move relevant modules into `trussed-core` crate
  - [x] https://github.com/trussed-dev/trussed/pull/178
- [x] review public dependencies
  - [x] `serde-indexed`
    - [x] https://github.com/trussed-dev/trussed/issues/183
  - [x] `heapless`
    - [x] https://github.com/trussed-dev/trussed/pull/186
- [x] review public API
  - [x] exhaustive enums
    - [x] https://github.com/trussed-dev/trussed/pull/189
  - [x] non-sealed traits
  - [x] module structure / re-exports
    - [x] https://github.com/trussed-dev/trussed/pull/190
  - [x] `Id::hex` etc.
    - [x] https://github.com/trussed-dev/trussed/pull/186
  - [x] `api::NotBefore` → `types::NotBefore`
    - [x] https://github.com/trussed-dev/trussed/pull/188
- [x] add feature flags
  - [x] clients
    - [x] https://github.com/trussed-dev/trussed/pull/185
  - [x] mechanisms
    - [x] https://github.com/trussed-dev/trussed/issues/159
- [x] Update docs.rs config
  - [x] https://github.com/trussed-dev/trussed/pull/191
- [x] release

robin-nitrokey:
Done: https://github.com/trussed-dev/trussed/releases/tag/core-v0.1.0.

PR #178, opened 2024-10-31
Add trussed-core crate

robin-nitrokey:
This PR adds the trussed-core crate (https://github.com/trussed-dev/trussed/issues/177) and moves the relevant modules fully or partially to it:
- `api`
- `client` (only `*Client` and mechanism traits as well as error and result types)
- `config` (only relevant constants)
- `error`
- `interrupt`
- `serde_extensions` (only client traits and error and result types)
- `types` (only those types used by other modules)

All moved types are re-exported from their old locations, so this is a compatible change for Trussed’s dependents.  As the extension request and reply serialization and deserialization is now split between `trussed-core` and `trussed`, I’ve added serialization and deserialization methods to the `Extension` trait so that `trussed` does not need to make assumptions about the serialization mechanism used by `trussed-core`.

We will probably want to make some more adjustments before releasing.  This PR only intends to perform the split with minimal modifications.


sosthene-nitrokey (approved):
Right, and we might want to merge before https://github.com/trussed-dev/trussed/pull/181 and https://github.com/trussed-dev/trussed/pull/182

sosthene-nitrokey:
A shame that because the original files are trimmed and not fully deleted, it doesn't show as a rename in the diff.

I'm trusting the line count that nothing is missing.

robin-nitrokey:
I’ve also tested this against nitrokey-3-firmware so I’m pretty sure that nothing is missing.  But every rebase risks losing something. :(  I’ll merge it now – updating the other PRs is hopefully trivial because `store` is not touched by this change.

sosthene-nitrokey:
Yes, updating these PRs will be much easier than rebasing this one.

PR #181, opened 2024-12-11
Delete all empty parent directories if they are empty

sosthene-nitrokey:
This saves a lot of memory since each directory takes at least one block We are starting to see this as an issue because the filesystems are now much deeper due to the SE050

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -72,7 +72,7 @@
 //! - Alternative: subdirectory <==> RP hash, everything else in flat files
 //! - In any case need to "list dirs excluding . and .." or similar
 
-use littlefs2::{driver::Storage, fs::Filesystem};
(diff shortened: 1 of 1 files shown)


robin-nitrokey (commented):
I think this would break existing code that relies on the fact that the parent directory always exists if a file was deleted successfully, for example:

https://github.com/Nitrokey/fido-authenticator/blob/63a14793877f49e0bd6a99d28834a0013fdb9d64/src/ctap2.rs#L1856

sosthene-nitrokey:
I'll limit this to the VFS for now. In the future we can make a test release with it enabled for all filesystem.

daringer (approved):

robin-nitrokey (approved):

Issue #183, opened 2024-12-12
Remove serde implementations for API requests and replies


robin-nitrokey:
The request and reply structs in the `api` module use `serde-indexed` to implement `serde::{Deserialize, Serialize}`.  This is used both inside and outside of Trussed, for example:
- `Chacha8Poly1305::wrap_key` in Trussed:
  https://github.com/trussed-dev/trussed/blob/20c0ddef6139160d707cffd10e8d563f2cc25b54/src/mechanisms/chacha8poly1305.rs#L191-L194
- `CredentialId` in fido-authenticator:
  https://github.com/Nitrokey/fido-authenticator/blob/63a14793877f49e0bd6a99d28834a0013fdb9d64/src/credential.rs#L50-L56

This is leaking implementation details and a potential footgun as reordering fields in Trussed’s syscall structs could break credential IDs for fido-authenticator.  So I would propose to remove these implementations and instead use custom structs in those places.  As far as I see, only `reply::Encrypt` is used which only has three fields, so the overhead should be negligible.

Are you okay with that?

PR #184, opened 2024-12-16
Remove serde trait implementations for requests and replies

robin-nitrokey:
Implementing Serialize and Deserialize for the request and reply structs leaks implementation details, especially when using serde_indexed.  This patch removes these implementations for the request and reply structs and also for some other types that presumably only had them because they were used in these structs and that are not serialized or deserialized anywhere in the Trussed ecosystem.

Fixes: https://github.com/trussed-dev/trussed/issues/183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `nonce` argument to `wrap_key` and `unwrap_key` syscalls.
 - Use nonce as IV for Aes256Cbc mechanism.
 - Updated `cbor-smol` to 0.5.0.
+- Removed `serde::{Deserialize, Serialize}` implementations for the API request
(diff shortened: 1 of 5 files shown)


sosthene-nitrokey (approved):

sosthene-nitrokey:
Seing the need to implement this in two third-party crates, could this lead to an increase of binary size?

robin-nitrokey:
Yes, in my tests this led to an increase of ~ 500 bytes.  I thought that would just be acceptable.  But we could also add `WrappedKey` to `trussed::types`.

robin-nitrokey:
But as we need the type in Trussed anyway (for the `CryptoClient::wrap_key` implementation), we can also just make it public.

robin-nitrokey:
I’ve added `EncryptedData` to `trussed_core::types`.  An additional benefit is that it can directly implement the conversion from `reply::Encrypt` and to `request::Decrypt`.

PR #188, opened 2024-12-19
core: Move api::NotBefore into types

robin-nitrokey:
All other types that are used in the API are in `types`, so this patch also moves `api::NotBefore` there.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,0 +1,2 @@
+pub use trussed_core::api::*;
+pub use trussed_core::types::NotBefore;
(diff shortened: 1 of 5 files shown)


src/api.rs
```diff
@@ -0,0 +1,2 @@
+pub use trussed_core::api::*;
+pub use trussed_core::types::NotBefore;
```
sosthene-nitrokey, line 2: nit: I don't think we need to add the pub use here. We can just change the import in fido-authenticator.

sosthene-nitrokey (approved):

src/api.rs
robin-nitrokey: To make the migration as easy as possible, I tried to keep the changes compatible.  I would suggest to merge it like this and once everything is migrated to trussed-core, we can do a cleanup round on trussed itself.

PR #192, opened 2024-12-20
Fix compile-time mechanism checks

robin-nitrokey:
This PR adds missing variants to `trussed::types::IMPLEMENTED_MECHANISMS` and makes it easier to validate the selected mechanisms at compile time.

### DIFF
--- a/core/src/types.rs
+++ b/core/src/types.rs
@@ -452,163 +452,149 @@ impl Client {
     ];
 }
 
-#[derive(Copy, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
-#[non_exhaustive]
-pub enum Mechanism {
-    #[cfg(feature = "aes256-cbc")]
-    Aes256Cbc,
-    #[cfg(feature = "chacha8-poly1305")]
-    Chacha8Poly1305,
-    #[cfg(feature = "ed255")]
-    Ed255,
-    #[cfg(feature = "hmac-blake2s")]
-    HmacBlake2s,
-    #[cfg(feature = "hmac-sha1")]
-    HmacSha1,
-    #[cfg(feature = "hmac-sha256")]
-    HmacSha256,
-    #[cfg(feature = "hmac-sha512")]
-    HmacSha512,
-    // P256XSha256,
-    #[cfg(feature = "p256")]
-    P256,
-    #[cfg(feature = "p256")]
-    P256Prehashed,
-    #[cfg(feature = "p384")]
-    P384,
-    #[cfg(feature = "p384")]
-    P384Prehashed,
-    #[cfg(feature = "p521")]
-    P521,
-    #[cfg(feature = "p521")]
-    P521Prehashed,
-    #[cfg(feature = "brainpoolp256r1")]
-    BrainpoolP256R1,
-    #[cfg(feature = "brainpoolp256r1")]
-    BrainpoolP256R1Prehashed,
-    #[cfg(feature = "brainpoolp384r1")]
-    BrainpoolP384R1,
-    #[cfg(feature = "brainpoolp384r1")]
-    BrainpoolP384R1Prehashed,
-    #[cfg(feature = "brainpoolp512r1")]
-    BrainpoolP512R1,
-    #[cfg(feature = "brainpoolp512r1")]
-    BrainpoolP512R1Prehashed,
-    #[cfg(feature = "secp256k1")]
-    Secp256k1,
-    #[cfg(feature = "secp256k1")]
-    Secp256k1Prehashed,
-    // clients can also do hashing by themselves
-    #[cfg(feature = "sha256")]
-    Sha256,
-    #[cfg(feature = "tdes")]
-    Tdes,
-    #[cfg(feature = "totp")]
-    Totp,
-    #[cfg(feature = "trng")]
-    Trng,
-    #[cfg(feature = "x255")]
-    X255,
-    /// Used to serialize the output of a diffie-hellman
-    #[cfg(feature = "shared-secret")]
-    SharedSecret,
+macro_rules! generate_mechanism {
+    (
+        $(#[$outer:meta])*
+        $vis:vis enum $name:ident {
+            $(
+                #[cfg($($cfg_cond:tt)*)]
+                $(#[$inner:meta])*
+                $variant:ident,
+            )*
+        }
+    ) => {
+        $(#[$outer])*
+        $vis enum $name {
+            $(
+                #[cfg($($cfg_cond)*)]
+                $(#[$inner])*
+                $variant,
+            )*
+        }
 
-    /// Exposes the Raw RSA encryption/decryption primitive. Be aware this is dangerous.
-    /// Not having any padding can allow an attacker to obtain plaintexts and forge signatures.
-    /// It should only be used if absolutely necessary.
-    #[cfg(feature = "rsa2048")]
-    Rsa2048Raw,
-    /// Exposes the Raw RSA encryption/decryption primitive. Be aware this is dangerous.
-    /// Not having any padding can allow an attacker to obtain plaintexts and forge signatures.
-    /// It should only be used if absolutely necessary.
-    #[cfg(feature = "rsa3072")]
-    Rsa3072Raw,
-    /// Exposes the Raw RSA encryption/decryption primitive. Be aware this is dangerous.
-    /// Not having any padding can allow an attacker to obtain plaintexts and forge signatures.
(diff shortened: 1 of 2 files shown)


core/src/types.rs
```diff
@@ -609,6 +610,77 @@ impl Mechanism {
...
+            #[cfg(feature = "rsa4096")]
+            Self::Rsa4096Raw => panic!("missing mechanism: Rsa4096Raw"),
+            #[cfg(feature = "rsa2048")]
+            Self::Rsa2048Pkcs1v15 => panic!("missing mechanism: Rsa2048Pkcs1v15"),
+            #[cfg(feature = "rsa3072")]
+            Self::Rsa3072Pkcs1v15 => panic!("missing mechanism: Rsa3096Pkcs1v15"),
+            #[cfg(feature = "rsa4096")]
+            Self::Rsa4096Pkcs1v15 => panic!("missing mechanism: Rsa4096Pkcs1v15"),
+        }
+    }
```
sosthene-nitrokey, line 683: What is that for?

robin-nitrokey: So that we can give an indication which mechanism is missing, see: https://github.com/Nitrokey/nitrokey-3-firmware/blob/e6132a0b099f8c2b1b575b9c1f1252d112e1d908/components/apps/src/lib.rs#L480

Due to the lack of const formatting, we could otherwise just indicate that there is *some* mechanism missing which can be annoying to debug.  Or am I missing some other trick to get a better error message?

robin-nitrokey:
Updated to use a macro to generate the `Mechanism` enum and to provide a `const_eq` function instead of `#[repr(u8)]` (because that does not work with empty enums).

core/src/types.rs
sosthene-nitrokey: Oh, I see.

Can't we instead not have a bunch of const asserts?

Something like:

```rust
const _: () = {
    #[cfg(feature = "se050")]
    assert!(const_contains(mechanisms::ENABLED, Mechanism::BrainpoolP256R1)),
    …
};
```

Where `const_contains` is a `const` function that checks that the mechanism in second argument is contained in the first argument.

AFAIK the compile-time error message should show the line of which assert failed.

robin-nitrokey: No, not really.  The check is the other way round.  I don’t want to check the presence of an expected mechanism, I want to check that there is no unexpected mechanism that is enabled but not implemented by a backend or Trussed itself.

Consider the following scenario:  We start to use the new `HmacSha512` mechanism in an application which is currently not enabled in the runner.  This would lead to a runtime error.  So the idea of the check is to list all mechanisms that are activated in `trussed-core` and can be used by applications (`trussed-core::types::Mechanism::ENABLED`) and compare them with the mechanisms that are enabled in `trussed` itself (`trussed::types::IMPLEMENTED_MECHANISMS`).  The list of mechanisms that you see in the linked code are the mechanisms that don’t need to be implemented by `trussed` because they are implemented by other backends, so it’s a whitelist and not the list of checked mechanisms.

core/src/types.rs
```diff
@@ -452,163 +452,115 @@ impl Client {
...
+        "rsa3072": Rsa3072Raw,
+        /// Exposes the Raw RSA encryption/decryption primitive. Be aware this is dangerous.
+        /// Not having any padding can allow an attacker to obtain plaintexts and forge signatures.
+        /// It should only be used if absolutely necessary.
+        "rsa4096": Rsa4096Raw,
+
+        "rsa2048": Rsa2048Pkcs1v15,
+        "rsa3072": Rsa3072Pkcs1v15,
+        "rsa4096": Rsa4096Pkcs1v15,
+    }
```
sosthene-nitrokey, line 563: I would instead keep the real syntax for enums and have the macro parse `#[cfg(feature = "...")]` I believe this will be more understandable at a glance.

sosthene-nitrokey (changes requested):

core/src/types.rs
sosthene-nitrokey: Oh, I see.

core/src/types.rs
robin-nitrokey: I had this initially but then you can no longer use ` $(#[$inner:meta])*` to match the doc comment and potentially other attributes.  Maybe it would be possible to hard-code the `doc` attribute but I preferred this simpler solution that also handles all other attributes.

sosthene-nitrokey: This works:

```rust
macro_rules! generate_mechanism {
    (
        $(#[$outer:meta])*
        $vis:vis enum $name:ident {
            $(
                #[cfg($($cfg_cond:tt)*)]
                $(#[$inner:meta])*
                $variant:ident,
            )*
        }
    ) => { ... }
}
```

You then just need in the definition to have the `#[cfg(..)]` be before the doc comments:

```rust
generate_mechanism! {
    #[derive(Copy, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
    #[non_exhaustive]
    pub enum Mechanism {
        #[cfg(feature = "aes256-cbc")]
        Aes256Cbc,
        #[cfg(feature = "chacha8-poly1305")]
        Chacha8Poly1305,
        #[cfg(feature = "ed255")]
        Ed255,
        #[cfg(feature = "hmac-blake2s")]
        HmacBlake2s,
        #[cfg(feature = "hmac-sha1")]
        HmacSha1,
        #[cfg(feature = "hmac-sha256")]
        HmacSha256,
        #[cfg(feature = "hmac-sha512")]
        HmacSha512,
        // P256XSha256,
        #[cfg(feature = "p256")]
        P256,
        #[cfg(feature = "p256")]
        P256Prehashed,
        #[cfg(feature = "p384")]
        P384,
        #[cfg(feature = "p384")]
        P384Prehashed,
        #[cfg(feature = "p521")]
        P521,
        #[cfg(feature = "p521")]
        P521Prehashed,
        #[cfg(feature = "brainpoolp256r1")]
        BrainpoolP256R1,
        #[cfg(feature = "brainpoolp256r1")]
        BrainpoolP256R1Prehashed,
        #[cfg(feature = "brainpoolp384r1")]
        BrainpoolP384R1,
        #[cfg(feature = "brainpoolp384r1")]
        BrainpoolP384R1Prehashed,
        #[cfg(feature = "brainpoolp512r1")]
        BrainpoolP512R1,
        #[cfg(feature = "brainpoolp512r1")]
        BrainpoolP512R1Prehashed,
        #[cfg(feature = "secp256k1")]
        Secp256k1,
        #[cfg(feature = "secp256k1")]
        Secp256k1Prehashed,
        // clients can also do hashing by themselves
        #[cfg(feature = "sha256")]
        Sha256,
        #[cfg(feature = "tdes")]
        Tdes,
        #[cfg(feature = "totp")]
        Totp,
        #[cfg(feature = "trng")]
        Trng,
        #[cfg(feature = "x255")]
        X255,
        #[cfg(feature = "shared-secret")]
        /// Used to serialize the output of a diffie-hellman
        SharedSecret,

        #[cfg(feature = "rsa2048")]
        /// Exposes the Raw RSA encryption/decryption primitive. Be aware this is dangerous.
        /// Not having any padding can allow an attacker to obtain plaintexts and forge signatures.
        /// It should only be used if absolutely necessary.
        Rsa2048Raw,
        #[cfg(feature = "rsa3072")]
        /// Exposes the Raw RSA encryption/decryption primitive. Be aware this is dangerous.
        /// Not having any padding can allow an attacker to obtain plaintexts and forge signatures.
        /// It should only be used if absolutely necessary.
        Rsa3072Raw,
        #[cfg(feature = "rsa4096")]
        /// Exposes the Raw RSA encryption/decryption primitive. Be aware this is dangerous.
        /// Not having any padding can allow an attacker to obtain plaintexts and forge signatures.
        /// It should only be used if absolutely necessary.
        Rsa4096Raw,

        #[cfg(feature = "rsa2048")]
        Rsa2048Pkcs1v15,
        #[cfg(feature = "rsa3072")]
        Rsa3072Pkcs1v15,
        #[cfg(feature = "rsa4096")]
        Rsa4096Pkcs1v15,
    }
}
```

robin-nitrokey: Thanks!  I think I tried this but got an error due to an ambiguous rule.  Maybe I made the `cfg` optional?

sosthene-nitrokey: Yes making the cfg optional (or putting it at the end) creates ambiguity that the macro matcher cannot handle. A bit frustrating.

robin-nitrokey: Updated.

sosthene-nitrokey (approved):

PR #196, opened 2025-01-24
Simplify virtual store 

robin-nitrokey:
This patch uses the simplified Store trait to also simplify the store
implementation used by the virt module:  Instead of using references to
static storages and filesystems protected by a mutex, we can now just
use separate non-static instances for every test.

The only downside is that we can no longer provide raw access to the
IFS via the StoreProvider::ifs function, so we cannot run the
provisioner app with this platform implementation.  But I think this is
an acceptable tradeoff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for non-static channels:
   - Added lifetimes to `ClientImplementation` and `ServiceEndpoints`.
   - Added the `pipe::TrussedChannel` type.
+- Refactored the `Store` trait:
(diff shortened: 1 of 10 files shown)


robin-nitrokey:
With https://github.com/trussed-dev/littlefs2/pull/98, we can use `Box<dyn DynStorageAlloc>` instead of type parameters for the three storage types.  So I’ve removed the `StoreProvider` trait.  Callers can either use `StoreConfig::ram()` to quickly setup in-memory storage or compose their own `StoreConfig` (e. g. in the usbip runner).  As the store setup is now so easy and we use custom wrapper functions in most crates anyway, I’ve also removed the `with_{fs,ram}_client{s,}` helper functions to keep things simple.

sosthene-nitrokey (approved):

PR #200, opened 2025-02-26
Post-Quantum Cryptography

KyleKotowick:


src/key.rs
```diff
@@ -6,12 +6,12 @@ use serde::{de::Visitor, ser::SerializeMap, Deserialize, Serialize};
...
-    config::{MAX_KEY_MATERIAL_LENGTH, MAX_SERIALIZED_KEY_LENGTH},
-    Error,
-};
+use crate::Error;
+use trussed_core::config::MAX_SERIALIZED_KEY_LENGTH;
 
-pub type Material = Vec<u8, { MAX_KEY_MATERIAL_LENGTH }>;
+// Keys are often stored in serialized format (e.g. PKCS#8 used by the RSA backend),
+// so material max length must be serialized max length.
+pub type Material = Vec<u8, { MAX_SERIALIZED_KEY_LENGTH }>;
```
sosthene-nitrokey, line 14: This should not be changed, max `KEY_MATERIAL_LENGTH` should reflect the proper length for key material data.

src/store/keystore.rs
```diff
@@ -181,7 +181,7 @@ impl<S: Store> Keystore for ClientKeystore<S> {
 
         let location = self.location(secrecy, id).ok_or(Error::NoSuchKey)?;
 
-        let bytes: Bytes<{ MAX_KEY_MATERIAL_LENGTH }> = store::read(&self.store, location, &path)?;
+        let bytes: Bytes<{ MAX_SERIALIZED_KEY_LENGTH }> = store::read(&self.store, location, &path)?;
```
sosthene-nitrokey, line 184: Same here.

core/Cargo.toml
```diff
@@ -10,13 +10,17 @@ license.workspace = true
 repository.workspace = true
 
 [dependencies]
+cfg-if = "1.0"
```
sosthene-nitrokey, line 13: nit: I think in this case we could avoid the dependency. It's possible to do the following. `cfg_if` is fine it's just an FYI.

```rust
pub const MAX_SIGNATURE_LENGTH: usize = if cfg!(feature = "larger-signature") {
    10_000_000
} else {
    1_000
};
```

sosthene-nitrokey (changes requested):

Cargo.toml
```diff
@@ -56,6 +56,9 @@ hmac = "0.12"
 sha-1 = { version = "0.10", default-features = false, optional = true }
 sha2 = { version = "0.10", default-features = false }
 
+# PQC
+pqcrypto-mldsa = { version = "0.1.0", optional = true }
+
```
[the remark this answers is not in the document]
KyleKotowick: By "inline", do you just mean copying the value out of the `pqcrypto-mldsa` dependency and reproducing it as a constant in `trussed`?

One of the most challenging and frustrating aspects of working on this project was the extensive duplication of constants and magic numbers across the many crates. Note that the `pqcrypto-mldsa` dependency here in `trussed` is optional, and only happens if the PQC backend is enabled (in which case, the `pqcrypto-mldsa` dependency is required anyways). Therefore, having it present offers the benefit of reducing duplication of constants and magic numbers, while not adding an additional dependency since it will only be used if it's necessary anyways.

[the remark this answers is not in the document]
robin-nitrokey: Before I weigh in on the details, could you, @KyleKotowick, please clarify if the signature size is an implementation detail of the pqcrypto library or if it is a property of the standardized algorithm?  Is it likely to change in the future?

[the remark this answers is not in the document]
KyleKotowick: It's a property of the standardized algorithm. Since the ML-DSA standard is now finalized, it shouldn't change.

core/Cargo.toml
KyleKotowick: The `cfg-if` is required because your proposed solution requires any referenced crates within the `if` block to exist. Since I'm using it to conditionally reference the `pqcrypto_mldsa` crate, if I used your proposed approach, that crate will always have to be included as a dependency in order for it to compile.

src/key.rs
KyleKotowick: There seems to be an issue with the `trussed-rsa-backend` and `fido-authenticator` where it's using the wrong sizes for byte arrays. If you clone this repo, patch to use the local copy, and revert this (and the other reference) back to `MAX_KEY_MATERIAL_LENGTH`, you'll get all sorts of build errors when building `nitrokey-3-firmware/runners/usbip`.

`cargo build --features provisioner`
```
error[E0308]: mismatched types
   --> /mnt/c/Users/KyleKotowick/Documents/GitHub/sandbox-quantum/fido-authenticator/src/ctap1.rs:150:13
    |
145 |         Ok(register::Response::new(
    |            ----------------------- arguments to this function are incorrect
...
150 |             cert,
    |             ^^^^ expected `1024`, found `2258`
    |
    = note: expected struct `heapless_bytes::Bytes<1024>`
               found struct `heapless_bytes::Bytes<2258>`
```

`cargo build --features provisioner,backend-mldsa44,backend-mldsa65,backend-mldsa87`
```
error[E0308]: `?` operator has incompatible types
  --> /mnt/c/Users/KyleKotowick/Documents/GitHub/sandbox-quantum/trussed-rsa-backend/src/crypto_traits.rs:74:29
   |
74 |               serialized_key: key_parts.serialize().map_err(|_err| {
   |  _____________________________^
75 | |                 error!("Failed to serialize key parts: {:?}", _err);
76 | |                 ClientError::DataTooLarge
77 | |             })?,
   | |_______________^ expected `7523`, found `7519`
   |
   = note: `?` operator cannot convert from `heapless_bytes::Bytes<7519>` to `heapless_bytes::Bytes<7523>`
   = note: expected struct `heapless_bytes::Bytes<7523>`
              found struct `heapless_bytes::Bytes<7519>`

error[E0308]: mismatched types
   --> /mnt/c/Users/KyleKotowick/Documents/GitHub/sandbox-quantum/trussed-rsa-backend/src/lib.rs:215:30
    |
215 |     Ok(reply::SerializeKey { serialized_key })
    |                              ^^^^^^^^^^^^^^ expected `7523`, found `7519`
    |
    = note: expected struct `heapless_bytes::Bytes<7523>`
               found struct `heapless_bytes::Bytes<7519>`
```

Among others. I've tried to dig into this, but given that the mismatch for the last example is between 7519 bytes and 7523 bytes, which correspond to the `MAX_KEY_MATERIAL_LENGTH` and `MAX_SERIALIZED_KEY_LENGTH` sizes respectively, there's something fishy going on. If you could try it, you might have more luck identifying where the issue is.

sosthene-nitrokey: Yeah, that looks like the issue of using type aliases instead of newtypes, which means that there doesn't need to be explicit conversions between types.

I'm looking into it.

- https://github.com/Nitrokey/fido-authenticator/issues/126
- https://github.com/trussed-dev/trussed-rsa-backend/issues/20

Cargo.toml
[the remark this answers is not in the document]
robin-nitrokey: Thanks.  In that case, I don’t really see the benefit of using the dependency.  I understand that it is frustrating to update these constants in multiple crates, but the only proper solution for this specific problem would be to make the Trussed API generic over the message sizes and let the runner pick suitable values for their configuration.  This is not feasible at the moment – we would need one const generic type parameter per size constant.  In the medium term, we can hopefully solve this using a trait with associated constants or a custom const-generic type, but this is not possible in Rust at the moment.

But using a dependency for the PQC sizes does not solve this problem.  If you were to add a new mechanism, you would still need to update the message sizes manually.  And as the PQC sizes are standardized and not subject to change, I don’t see the benefits of pulling in a dependency for it.

core/src/config.rs
```diff
@@ -13,3 +9,54 @@ pub const SERDE_EXTENSION_REQUEST_LENGTH: usize =
...
+        MAX_SERIALIZED_KEY_LENGTH + 57
+    } else {
+        // Default from before addition of PQC
+        128
+    };
+
+// 30 bytes are added by CBOR serialization of a FullCredential
+// TODO: This was calculated by debugging and finding each location where this variable needed to be larger for one reason or another.
+// Update this to use different consts for each area where this is needed, instead of one const used everywhere.
+pub const MAX_MESSAGE_LENGTH: usize = MAX_FIDO_WRAPPED_KEY_LENGTH + 30 + 2031 + 32 + 37;
```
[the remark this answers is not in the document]
sosthene-nitrokey: You can use `cargo t -- --ignored` to run the test in `tests/interchange_size` to keep track of how high you bumped the interchange size.

sosthene-nitrokey (changes requested):

Issue #209, opened 2025-05-07
Review padding used for Aes256Cbc mechanism


robin-nitrokey:
Currently, the Aes256Cbc always uses zero padding:

https://github.com/trussed-dev/trussed/blob/024e0eca5fb7dbd2457831f7c7bffe4341e08775/src/mechanisms/aes256cbc.rs#L57

This is a counter-intuitive default.  It should either be configurable or default to no padding.

robin-nitrokey:
It looks like all applications that currently use Aes256Cbc provide data in multiples of the block size (to be confirmed) so it would be the best to use no padding in Trussed.

PR #210, opened 2025-05-13
aes256cbc: Use no padding instead of zero padding

robin-nitrokey:
Defaulting to zero padding can be problematic.  Using no padding leaves it up to the application to handle the padding correctly according to the use case.

Fixes: https://github.com/trussed-dev/trussed/issues/209

To do:  Review all applications and backends to ensure that only full blocks are used with AES-256.

### DIFF
--- a/src/mechanisms/aes256cbc.rs
+++ b/src/mechanisms/aes256cbc.rs
@@ -15,10 +15,9 @@ impl MechanismImpl for super::Aes256Cbc {
         request: &request::Encrypt,
     ) -> Result<reply::Encrypt, Error> {
         use aes::Aes256;
-        use cbc::cipher::{block_padding::ZeroPadding, BlockEncryptMut, KeyIvInit};
+        use cbc::cipher::{block_padding::NoPadding, BlockEncryptMut, KeyIvInit};
 
         type Aes256CbcEnc = cbc::Encryptor<Aes256>;
-        // TODO: perhaps use NoPadding and have client pad, to emphasize spec-conformance?
 
         let key_id = request.key;
         let key = keystore.load_key(key::Secrecy::Secret, None, &key_id)?;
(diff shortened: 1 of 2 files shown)


src/mechanisms/aes256cbc.rs
```diff
@@ -51,10 +50,8 @@ impl MechanismImpl for super::Aes256Cbc {
         // hprintln!(" aes256cbc encrypting l = {}B: {:?}", l, &buffer).ok();
 
         // Encrypt message in-place.
-        // &buffer[..pos] is used as a message and &buffer[pos..] as a reserved space for padding.
-        // The padding space should be big enough for padding, otherwise method will return Err(BlockModeError).
         let ciphertext = cipher
-            .encrypt_padded_mut::<ZeroPadding>(&mut buffer, l)
+            .encrypt_padded_mut::<NoPadding>(&mut buffer, l)
```
sosthene-nitrokey, line 54: Maybe don't panic if the padding is wrong, and error instead

src/mechanisms/aes256cbc.rs
```diff
@@ -140,12 +136,11 @@ impl MechanismImpl for super::Aes256Cbc {
...
 
         // Decrypt message in-place.
-        // Returns an error if buffer length is not multiple of block size and
-        // if after decoding message has malformed padding.
+        // Returns an error if buffer length is not multiple of block size
         // hprintln!("encrypted: {:?}", &buffer).ok();
         // hprintln!("symmetric key: {:?}", &symmetric_key).ok();
         let plaintext = cipher
-            .decrypt_padded_mut::<ZeroPadding>(&mut buffer)
+            .decrypt_padded_mut::<NoPadding>(&mut buffer)
```
sosthene-nitrokey, line 143: Same, error instead of panicking.

sosthene-nitrokey (commented):
Reviewed applications:

- [x] Opcard OK (we check length and reject non multiple of 16 bytes)
- [x] PIV: AES only ever encrypts/decrypts 16 bytes (and check the length)
- [ ] FIDO
- [ ] Secrets

src/mechanisms/aes256cbc.rs
robin-nitrokey: Right, fixed.

robin-nitrokey:
There is actually one case in fido-authenticator where zero-padding is performed: the encrypted PIN when setting a new PIN.  I’ll prepare a PR with this change.

robin-nitrokey:
Nevermind, fido-authenticator already strips the padding (though this was probably not necessary with the current implementation):

https://github.com/Nitrokey/fido-authenticator/blob/5ebb4a48302e4c80a2abe1c2d86201a3df2a1d2d/src/ctap2.rs#L1247

sosthene-nitrokey (approved):

PR #219, opened 2026-05-10
Add mldsa44 mechanism implementation

0x0ece:
Two changes:
1. We now need a press and long-press events, so we can support the proper reset according to specs (long-press >= 5s, instead of our custom plug-pressing)
2. ml-dsa-44, using rust-crypto, it should work on m4/m33. I saw the other PR with definitions for all post quantum algos, I used the same def for features, realistically we won't do more than 128-bit security so to me it doesn't make sense to have code for all 3, but we can add later if needed.

### DIFF
--- a/src/mechanisms/mldsa44.rs
+++ b/src/mechanisms/mldsa44.rs
@@ -1,0 +1,259 @@
+use libcrux_ml_dsa::ml_dsa_44::{
+    generate_key_pair, sign, verify, MLDSA44Signature, MLDSA44VerificationKey,
+};
+use libcrux_ml_dsa::{KEY_GENERATION_RANDOMNESS_SIZE, SIGNING_RANDOMNESS_SIZE};
+use rand_core::RngCore;
+
+use crate::api::{reply, request};
+use crate::error::Error;
+use crate::key;
+use crate::service::MechanismImpl;
+use crate::store::keystore::Keystore;
+use crate::types::{KeyId, KeySerialization, SerializedKey, Signature, SignatureSerialization};
+
+const SEED_LEN: usize = KEY_GENERATION_RANDOMNESS_SIZE; // 32
+const PUBLIC_KEY_LEN: usize = 1312;
+const SIGNATURE_LEN: usize = 2420;
+
+#[inline(never)]
+fn load_seed(keystore: &mut impl Keystore, key_id: &KeyId) -> Result<[u8; SEED_LEN], Error> {
+    let seed: [u8; SEED_LEN] = keystore
+        .load_key(key::Secrecy::Secret, Some(key::Kind::Mldsa44Seed), key_id)?
+        .material
+        .as_slice()
+        .try_into()
+        .map_err(|_| Error::InternalError)?;
+    Ok(seed)
+}
+
+#[inline(never)]
+fn load_public_encoded(
+    keystore: &mut impl Keystore,
+    key_id: &KeyId,
+) -> Result<crate::Bytes<{ crate::config::MAX_KEY_MATERIAL_LENGTH }>, Error> {
+    let key = keystore.load_key(key::Secrecy::Public, Some(key::Kind::Mldsa44Seed), key_id)?;
+    Ok(key.material)
+}
+
+// Non-inline wrappers around libcrux so its sign/verify machinery
+// lives in its own stack frame (not folded into our Mldsa44::sign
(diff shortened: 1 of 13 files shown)


src/mechanisms/mldsa44.rs
```diff
@@ -0,0 +1,251 @@
...
+                // map keys and we need exact bytes regardless of crate quirks.
+                if material.len() != PUBLIC_KEY_LEN {
+                    return Err(Error::InternalError);
+                }
+                // Header bytes for the COSE map:
+                //   a3        — map(3)
+                //   01 07     — 1: 7 (kty = AKP)
+                //   03 38 31  — 3: -50 (alg = ML-DSA-44; -50 = ~49 = 0x31)
+                //   20        — -1 (key = pub)
+                //   59 05 20  — bstr(1312)
```
robin-nitrokey: I find the comment “key = pub” somewhat misleading. I would put it like this:
```suggestion
                //   01 07     — key 1 = kty, value 7 = APK
                //   03 38 31  — key 3 = alg, value -50 = ML-DSA-44 (-50 = ~49 = 0x31)
                //   20        — key -1 = pub
                //   59 05 20  — bstr with length 1312 = 0x0520
```

src/mechanisms/mldsa44.rs
```diff
@@ -0,0 +1,251 @@
...
+                let header: [u8; 9] = [0xa3, 0x01, 0x07, 0x03, 0x38, 0x31, 0x20, 0x59, 0x05];
+                serialized_key
+                    .extend_from_slice(&header)
+                    .map_err(|_| Error::InternalError)?;
+                serialized_key
+                    .extend_from_slice(&[0x20])
+                    .map_err(|_| Error::InternalError)?; // low byte of 1312 (0x0520)
```
robin-nitrokey: Why is this not part of `header`?

robin-nitrokey: Maybe it would be clearer to use `PUBLIC_KEY_LEN` here instead of duplicating the magic constant?

src/mechanisms/mldsa44.rs
```diff
@@ -0,0 +1,251 @@
...
+                serialized_key
+                    .extend_from_slice(&material)
+                    .map_err(|_| Error::InternalError)?;
+            }
+            KeySerialization::Cose => {
+                // COSE_Key for ML-DSA-44 (draft-ietf-cose-dilithium /
+                // draft-ietf-cose-akp): map with kty=AKP(7), alg=-50, pub=bstr.
+                //   {1: 7, 3: -50, -1: <1312-byte public key>}
+                // Hand-encoded CBOR — `cbor-smol` is awkward for negative-int
+                // map keys and we need exact bytes regardless of crate quirks.
```
robin-nitrokey: Just out of curiosity, what problem did you have with cbor-smol?

robin-nitrokey (commented):
LGTM, I just have some small comments regarding the COSE serialization. Also I want to give @sosthene-nitrokey a chance to review before merging.

Previously we were more reluctant to add experimental features to trussed directly (and would maybe have used a custom backend instead), but with the increased separation due to trussed-core and as this is implemented using a rust-crypto crate, I think it is totally fine.

robin-nitrokey:
I’ve merged the changes to trussed-core separately so that this now only contains the mldsa44 implementation in trussed.
- https://github.com/trussed-dev/trussed/pull/220
- https://github.com/trussed-dev/trussed/pull/221

sosthene-nitrokey:
LibCrux appears to be using a global allocator which would be a new requirement for trussed. The RSA crate that uses an allocator too is separated into its own backend for this reason.

0x0ece:
> LibCrux appears to be using a global allocator which would be a new requirement for trussed

where do you see that? I looked at the code and asked claude but can't find it. To me it looks like it's all stack.

sosthene-nitrokey:
The [README](https://github.com/cryspen/libcrux#:~:text=libcrux%20and%20the%20individual%20primitive%20crates%20it%20depends%20on%20support%20no%5Fstd%20environments%20given%20a%20global%20allocator%20for%20the%20target%20platform) in the repo mentions it.

> libcrux and the individual primitive crates it depends on support no_std environments given a global allocator for the target platform.

0x0ece:
Maybe it's for some specific crates, AEAD seems to use alloc in some cases, but ml-dsa doesn't.

PR #224, opened 2026-05-30
Add Aes256Gcm mechanism and release trussed-core v0.2.2

robin-nitrokey:
This PR makes the Chacha8Poly1305 implementation generic over the AEAD algorithm, adds the Aes256Gcm mechanism to trussed-core and trussed using this generic implementation and releases trussed-core v0.2.2.

### DIFF
--- a/src/mechanisms/aead.rs
+++ b/src/mechanisms/aead.rs
@@ -1,0 +1,229 @@
+use core::marker::PhantomData;
+
+use aead::{generic_array::GenericArray, AeadCore, AeadMutInPlace};
+use cipher::{typenum::U32, ArrayLength, KeyInit, KeySizeUser, Unsigned as _};
+use rand_core::RngCore as _;
+use trussed_core::{
+    api::{reply, request},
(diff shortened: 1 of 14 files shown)


0x0ece (approved):

src/mechanisms/aead.rs
```diff
@@ -0,0 +1,229 @@
...
+use trussed_core::{
+    api::{reply, request},
+    types::{EncryptedData, Mechanism, Message, ShortData},
+    Error,
+};
+
+use crate::{key, store::keystore::Keystore};
+
+pub struct Aead<T, KeyNonceSize> {
+    mechanism: Mechanism,
```
sosthene-nitrokey, line 15: Why use a struct field instead of a trait `const`?

sosthene-nitrokey (approved):

src/mechanisms/aead.rs
robin-nitrokey: To which trait would you add it? Or create a new trait just for the field?

daringer (approved):

PR #226, opened 2026-06-02
Add test for invalid key size for AEAD mechanisms

robin-nitrokey:
This patch adds a test for the panic on an invalid key size for AEAD mechanisms that was reported by @MG3004 in GHSA-32mp-78p5-q55v and fixed in #224.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,6 +158,10 @@ test-attestation-cert-ids = []
 test-increased-interchange-size = []
 
 [[test]]
+name = "aead"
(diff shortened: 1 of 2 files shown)


sosthene-nitrokey (approved):
It would be good to also have some tests with encryption/decryption where the nonce is of the wrong length, especially through `unwrap_key` because this one could easily be called with user-provided data.

robin-nitrokey:
@sosthene-nitrokey Already prepared. :)

